PR #9, opened 2023-05-16
Basic Leveling run type, only for sorc

hectorgimenez:
- Basic Leveling (only Act 1 implemented)
- Sorc only, it's able to put stats and skill points based on requirements
- Return to town during pickup to sell/stash if the inventory is full
- Read belt rows from memory, we don't need to put it in the config anymore
- Added new action type that is able to chain multiple actions. At some point I need to refactor all this action thing...
- New action that is able to traverse map (by walking) while killing monsters, more or less, needs tweaking
- Some kind of waypoint autodiscover
- Reworked the build script
- Restored default keybindings, I use weird ones
- Fixed corpse detection and recover
- Changed a bit the exit game logic, hopefully it will work a bit better

I have no idea how to read questing related stuff from memory, same for waypoints. So I don't know how to check if a quest is already done or waypoint already discovered.


internal/action/clear_level.go
hectorgimenez, line 1: Check if this cause any problem with this action on normal (non leveling) usage.

PR #12, opened 2023-07-18
buy from Drognan

johnny-yue:
Lysander does not sell ID scroll, when it runs out the bot will stuck.

Click once at the start of a run to ensure it is not holding anything at hand. Otherwise it get stuck.

### DIFF
--- a/internal/action/leveling_tools.go
+++ b/internal/action/leveling_tools.go
@@ -27,6 +27,14 @@ var uiSkillTabPosition = []data.Position{
 var uiSkillRowPosition = [6]int{190, 250, 310, 365, 430, 490}
 var uiSkillColumnPosition = [3]int{920, 1010, 1095}
 
+
+func (b Builder) EnsureEmptyHand() *StaticAction {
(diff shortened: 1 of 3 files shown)


johnny-yue:
holding an item happens when she tries to ID an item, but ID tome is empty, so it turns into a "grab". I guess if buying ID scroll problem is taken care of, there is no reason for she to grab one item in hand.

hectorgimenez:
That's a very good point. I'm not 100% sure how to properly solve this problem, we have different options:

- Keep identifying items until we run out of IDs, then just stash those items unid
- Ensure that we have enough IDs before we start identifying (count number of unid items and if IDs is lower, go to vendor to refill)
- Use Cain (I don't like this option because Cain might be not available or maybe at some point we can add support for stashing some items unid).

johnny-yue:
I think ensuing ID scroll is bigger than 0 makes the most sense. 
Also really like the idea of identifying items immediately after picking up and throwing out junk without going back to town. This will save a lot of time going back and forth, and allow the player to carry more charm (less space to store junk)

PR #14, opened 2023-07-24
Improve countess for blizzard sorc

johnny-yue:
1. Remove last shot TP in front of monster. Set stop distance when moving to the box.

2. Cast FrostNova if there are many monster near player.

### DIFF
--- a/internal/action/leveling_tools.go
+++ b/internal/action/leveling_tools.go
@@ -27,6 +27,14 @@ var uiSkillTabPosition = []data.Position{
 var uiSkillRowPosition = [6]int{190, 250, 310, 365, 430, 490}
 var uiSkillColumnPosition = [3]int{920, 1010, 1095}
 
+
+func (b Builder) EnsureEmptyHand() *StaticAction {
+	return BuildStatic(func(d data.Data) (steps []step.Step) {
+		hid.Click(hid.LeftButton)
+		return nil
+	})
+}
+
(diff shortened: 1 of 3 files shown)


hectorgimenez:
Interesting, I never used frost nova with my blizzard build (I always go with act2 holy freeze merc). Did you only miss this in Countess?

Maybe it would be good to find a generic situation where we can use it, maybe even during the normal attack rotation (like the blizzard cast over own character when many mobs are very close). Just thinking about other cases like nihlathak, andariel, diablo seals... where many monsters can jump very fast over us. Do you think it will bring more survavility?

johnny-yue:
I ran some Andarial, countess, Mephisto, Nihlathak, and ancient tunnels so far. 
Countess should be quite safe but the char is having problems so I am trying to work on it first. For Nihlathak I do notice some monster jumps very quickly at the player and the player is not responding quickly enough. In ancient tunnel sometimes the player will stand in a pack of monster, but those are weak so no major problem.

My char has Act 2 holy freeze as well. Probably the Holy Freeze has an internal GCD. I notice it sometimes doesn't trigger at the time I would like. FrostNova is a sure thing to slow down nearby monsters.

I do think adding FrostNova to the generic attack rotation would be very nice.
Furthermore, I thought a priority-based attack rotation could be a good idea to increase kill speed.
Prioritizing super unique, unique, big pack, etc. Leave a few isolated mobs along. This could be quite good for 85lv areas like ancient tunnel.

PR #15, opened 2023-07-28
add runs, fix coordinate

johnny-yue:
1. After collision grid is expanded, coordinate was not properly translated back to game coordiate, causing cached path to be not found. So the engine must do parse world and path find at each iterateion

2. Use PlainKind when expanding collision grid, makes pathfider faster

3. add stony tomb, pit, arachnid lair (aka spider cave)

### DIFF
--- a/config/config.yaml.dist
+++ b/config/config.yaml.dist
@@ -78,11 +78,12 @@ character:
 
 game:
   clearTPArea: true # Will clear the TP area before clicking it
-  difficulty: hell  # Allowed values: normal, nightmare, hell
+  difficulty: hell # Allowed values: normal, nightmare, hell
   randomizeRuns: true
   # Just add the runs you want to do and they will be executed respecting the order, unless randomizeRuns is set to true
-  # Available runs: countess, andariel, ancient_tunnels, summoner, mephisto, council, eldritch, pindleskin, nihlathak, tristram, lower_kurast
-  runs: [ andariel, summoner, mephisto, pindleskin ]
+  # Available runs: countess, andariel, ancient_tunnels, summoner, mephisto, council, eldritch, pindleskin, nihlathak, 
+  #                 tristram, lower_kurast, stony_tomb, pit, arachnid_lair
+  runs: [stony_tomb, pit, arachnid_lair]
 
   # Specific runs settings
   pindleskin:
--- a/internal/action/step/move_to.go
+++ b/internal/action/step/move_to.go
@@ -96,7 +96,7 @@ func (m *MoveToStep) Run(d data.Data) error {
 
 	stuck := m.isPlayerStuck(d)
 
-	if m.path == nil || !m.cachePath(d) || stuck {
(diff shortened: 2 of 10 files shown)


hectorgimenez:
Hey @johnny-yue thanks for this! Could you consider changing the target branch to act2_leveling instead main? There are tons of changes there, including a change that will attach the target map to the current one instead of expanding the collision grid. It makes pather calculation much faster and more accurate (because it can consider destination map collision grid).

I plan to merge that branch to main soon when I have some free time, there is also a new movement action that allows characters to move without teleport and act 1-5 auto leveling, it doesn't work very well atm, but I also added some extra runs.

johnny-yue:
Sure thing. I was just looking at the act2_leveling branch. Struggled with it a bit to try to compile it. 
1. Had to change to a posix mingw binary, otherwise the compiler will complain about mutex related stuff when compiling opencv.
2. Modified the build script in gocv so that it builds static libraries. otherwise, linker will complain about missing library.

Might be useful to bring it up in the readme.

hectorgimenez:
Yeah... I added opencv to identify current quest status, skills, etc... just because I have no idea how to read those things from memory. So if we find the memory offsets where that information it is, we can drop the opencv thing. But I'm not good on that.

I also struggled with the same to compile opencv, yeah it might be good to update the readme. I don't have my computer with me, I will do it once I come back from vacations. Also it might be good if we can create a dockerfile with the required dependencies to generate the binary, I'm happy being able to compile on my computer because I use the debugger a lot, but  might help others.

johnny-yue:
Very interested to know what is the general approach to identify memory location for certain data, if you can share a bit.
Another thing, my debugger seems not able to pause all threads, so once a breakpoint is hit, the mouse cursor is still clicking, keyboard typing. I am using vs code and its the default debugger that comes with golang plugin.

hectorgimenez:
Honestly I'm not sure neither, all of those offsets are from other open sourced projects, I've been playing around with cheat engine and ghidra a bit, but I'm not an expert on that. I've been able to find those memory addresses but I have no idea about the process to follow to locate them again once they change when a new update is released. Same for map seed, this project is relying on this: https://github.com/joffreybesos/rustdecrypt that probably will not be updated in the future.

You can try to find more info at blizzhackers discord, is all the guidance that I can bring, I'm also interested on that part, so if you can get more info, feel free to share it :)

About the debugger, I use IntelliJ, but I think it must be pretty much the same since both are using delve. One thing I need to do in order to debug properly is to [comment the line that is registering the keyboard event hooks, on main.go](https://github.com/hectorgimenez/koolo/blob/act2_leveling/cmd/koolo/main.go#L137) because it's trying to handle those events in a different thread and debugger gets crazy.

Also keep in mind that Action and Steps is a hell, but all the callbacks are executed sequentially when their turn arrives, it's a bit messy. I was just thinking on making the data.Data a global variable and remove all the callback nesting hell, but I need to do some PoC.

Issue #75, opened 2024-02-14
Request: Implementation of a follower functionality


githubpersttn:
Hello,

Great work here!

Looking at the companion functionality it made me thinking why not to implement a "follower" functionality which will be a simplified version of the companion.

The use case is when you have D2R in e.g. virtual machines and you want your other accounts to follow you. This could be a party with you main character being a blizz sorc but other accounts are a summoner(s) and paladins.

The minimal implementation is just to follow your leader, changing the areas as well. A little bit more advanced could be joining the game, entering a portal, using a portal to enter a city. In the most advanced case an attack sequence can be added, but that's probably not important. As long as e.g. a summoner and a pala follows you it's enough.

This functionality will be useful for people who want to plain in a solo party or during new ladders.

Regards

hectorgimenez:
It's a good point and I will lie if I say I've been not thinking about it many times, it could be good to implement it but I don't think I have enough time for it. Let's keep it in the wishlist, for now I'm focusing on making it less invasive so we can run it without VM and keep using the computer. (or tu run multiple instances in one single VM)

13413j1j13j5315n13:
With such as massive change recently, this is not relevant per se anymore. Will open a new ticket later if needed

PR #76, opened 2024-02-18
#75 Add initial implementation of a follower

13413j1j13j5315n13:
This pull request adds a proof of concept implementation of a follower.

The logic is super basic. Just follow the leader if the leader is within a distance of 100. The distance of 100 means about one screen with the resolution of 1280x720. IMO more than enough.
If the leader has changed it's position (e.g. entered Den of Evil) the distance difference will be very high and thus a follower will not do anything anymore. Manual intervention to click on the entrance is needed.

I had to change the healing logic to check whether we're in town because with follower we will need to start it outside town.

Other than that, it's just an initial implementation and of course we can add more logic later on.

This implements the basic of https://github.com/hectorgimenez/koolo/issues/75

### DIFF
--- a/cmd/koolo/main.go
+++ b/cmd/koolo/main.go
@@ -90,7 +90,10 @@ func main() {
 
 	var supervisor koolo.Supervisor
 	var companion koolo.Companion
-	if config.Config.Companion.Enabled {
+	if config.Config.Follower.Enabled {
+		supervisor = koolo.NewFollowerPlayerSupervisor(logger, bot, gr, gm)
+		companion = supervisor.(koolo.Companion)
+	} else if config.Config.Companion.Enabled {
 		supervisor = koolo.NewCompanionSupervisor(logger, bot, gr, gm)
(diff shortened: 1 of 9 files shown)


hectorgimenez:
Hey, thanks for the contribution. I think maybe we can just reuse current Companion? I think what the companion is doing right now is very very basic and sharing same principles as Follower idea, so why not just overwriting current Companion with this? I really think it's a good starting point, and in the not so far future I believe it can be a very good complement for the leveling runs, so we can level multiple chars together.

13413j1j13j5315n13:
> Hey, thanks for the contribution. I think maybe we can just reuse current Companion? I think what the companion is doing right now is very very basic and sharing same principles as Follower idea, so why not just overwriting current Companion with this? I really think it's a good starting point, and in the not so far future I believe it can be a very good complement for the leveling runs, so we can level multiple chars together.

That makes sense.
I will check it.

PR #77, opened 2024-02-20
Enhance companion run with follower logic

13413j1j13j5315n13:
This tries to implement a proof of concept of follower using Companion run.

Some things are taken from https://github.com/hectorgimenez/koolo/pull/76

This should implement the basic of https://github.com/hectorgimenez/koolo/issues/75

I however do not have very high level characters for proper testing. However, we can enhance more if bugs will come

### DIFF
--- a/internal/run/companion.go
+++ b/internal/run/companion.go
@@ -1,15 +1,19 @@
 package run
 
 import (
-	"time"
-
+	"fmt"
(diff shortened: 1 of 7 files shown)


hectorgimenez:
Awesome, looks promising, I think we can use it as a base for a good party leveling sequence maybe

PR #101, opened 2024-03-21
Open a TP in baal run and other runs if we are a leader, use its own function to leader TP

13413j1j13j5315n13:


hectorgimenez:
Great, thanks!

PR #108, opened 2024-03-25
Make skill and stat functionality during leveling optional through config

13413j1j13j5315n13:
This pull request adds the possibility to disable skill, stat, reset, and key bindings functionality through config values.

This is done to have flexibility for someone who does not want to these functionalities enabled. E.g. knowing better what and howto allocate, having specific items which, set up the key bindings beforehand etc.

At least personally I've experienced sometimes bot messing up with key bindings and puts TP on F5 and armor on F4. It's easier to set everything myself once and forget about it, however make sure that these will not be reset again accidentally.

### DIFF
--- a/config/template/config.yaml
+++ b/config/template/config.yaml
@@ -92,6 +92,9 @@ game:
     clearArea: true
   baal:
     killBaal: false
+  leveling:
+    ensurePointsAllocation: true # Bot will allocate skill and stat points by itself or perform stat/skill reset. Set to false if you do NOT want it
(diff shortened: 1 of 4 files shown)


hectorgimenez:
This sounds very good, honestly I would like to spend some time working on the leveling tool and have some ideas, maybe at some point if I have some time after bugfixing. I will take a look at the PR tomorrow, but feel free to join the discord https://discord.gg/RWuQ78P7

PR #111, opened 2024-03-26
Add option to chicken on dolls and/or souls in baal run

hecuba:
This adds the option to exit the game immediately when souls or dolls are present in the throne room at baal


13413j1j13j5315n13:
I think this functionality would be nice to have.
But does it make sense to generalize this function into its own action (e.g. `chicken.go`?) and pass an array of monsters as a parameter to that function? This way we could use it in other places as well.

RobotGaston:
Good function but it's incompatible with a new main. Please fix this . : )

hecuba:
Updated it to the new data format.

I currently do not find the time to refine it or put it into a dedicated chicken tool file. 
Additionally, I can't think of another use case for a chicken on a specific enemy type. It might be rare enough to just keep this "hacky" implementation in the baal run script.

hectorgimenez:
Will close this one in favor of PR https://github.com/hectorgimenez/koolo/pull/148

Thanks.

PR #134, opened 2024-04-10
Add option to not mule into personal, but shared stash (hardcore)

theksk23:
Hola,
a little something we hardcore players desire, not quite sure if we should put that behind a generic `character: hardcore` setting, or have something more specific like `character.muleToShared`? 

Thanks!

### DIFF
--- a/internal/action/stash.go
+++ b/internal/action/stash.go
@@ -117,6 +117,9 @@ func (b *Builder) stashGold(d data.Data) {
 
 func (b *Builder) stashInventory(d data.Data, forceStash bool) {
 	currentTab := 1
+  if b.CharacterCfg.Character.Hardcore {
+    currentTab = 2
+  }
 	b.switchTab(currentTab)
 
 	itemsInStashTabs := slices.Concat(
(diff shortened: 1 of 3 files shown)


theksk23:
Looking at that in this graphical manner: tabs or spaces?

hectorgimenez:
Hola,

I think it's a good idea, I would say maybe better a `stashToShared` or something like this, since it will be more configurable? I assume `hardcore: true` in the future can contain also other stuff focused on hardcore playing, and maybe someone by any reason wants to stash on shared even if not playing hardcore, so I would prefer to keep things as specific as possible to cherrypick.

Gracias!

internal/action/stash.go
```diff
@@ -117,6 +117,9 @@ func (b *Builder) stashGold(d data.Data) {
 
 func (b *Builder) stashInventory(d data.Data, forceStash bool) {
 	currentTab := 1
+  if b.CharacterCfg.Character.StashToShared {
```
13413j1j13j5315n13, line 120: Did you run go fmt on this file?

theksk23:
me is a total Go noob,
Will look that up and run it, thanks for the hint!
(did disable Neovim autoformat because it was re-ordering imports, eh..)



13413j1j13j5315n13 ***@***.***> schrieb am Do. 11. Apr. 2024
um 11:32:

theksk23:
that indeed fixed the formatting, thanks!

Issue #145, opened 2024-04-11
Bot tries to open chests forever


NRHGDW:
### Details

![image](https://github.com/hectorgimenez/koolo/assets/99230615/77b5982f-b02d-45bf-9472-59016a50bd29)

Has been standing there trying for quite some time. He will continue until the game times out. If I buy keys, he sells them. If I add keys to nip he picks them up then sells them. I disabled everything chest-related in config but doesn't seem there's a global option; I'd rather skip them than waste a run. I don't know, what is the proper way to set up the use of keys for chests? Or is one coming? Thanks

### Version

v0.4.0 beta5

mista79:
Try placing the key in a locked area in the inventory.

nullptr-ref:
@mista79 that's all fine but won't help when the bot doesn't buy keys to start with and what happens when it runs out?

mista79:
I didn't have this problem after placing the key into the locked area.
The bot always buys keys and will not sell them because it is in a locked area.

nullptr-ref:
Be that as it may, the fact that the bot doesn't buy keys to start with is an issue as it will lead to never-ending loop of trying to open chests that are locked.

mista79:
Yes, it should check if the key is present and if not, won't open the box.

nullptr-ref:
This should be fixed with #152

artosimonyan:
This will be fixed in the upcoming versions.

PR #149, opened 2024-04-13
Light Sorc improvement, Kill process on stop and LK run improvements

nullptr-ref:
- Change Lightening sorc max range (to 10 from 15) - This improves the overall flow of the lightening sorc as it sometimes tries to attack mobs outside of the nova range.
- Add process kill on supervisor stop.
- Added a check for keys in inventory before trying to open locked chests in LK runs.  mentioned in #145 . We probably need to add some kind of max tries in the InteractObject logic to fix this globally.

### DIFF
--- a/internal/run/lower_kurast_chests.go
+++ b/internal/run/lower_kurast_chests.go
@@ -1,13 +1,15 @@
 package run
 
 import (
+	"slices"
+
 	"github.com/hectorgimenez/d2go/pkg/data"
 	"github.com/hectorgimenez/d2go/pkg/data/area"
+	"github.com/hectorgimenez/d2go/pkg/data/item"
 	"github.com/hectorgimenez/d2go/pkg/data/object"
 	"github.com/hectorgimenez/koolo/internal/action"
 	"github.com/hectorgimenez/koolo/internal/game"
 	"github.com/hectorgimenez/koolo/internal/pather"
-	"slices"
 )
 
 var bonfireName = object.SmallFire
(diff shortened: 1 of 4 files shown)


internal/run/lower_kurast_chests.go
```diff
@@ -55,24 +57,32 @@ func (a LowerKurastChest) BuildActions() []action.Action {
...
-									for _, obj := range d.Objects {
-										isSameObj := obj.Name == chest.Name && obj.Position.X == chest.Position.X && obj.Position.Y == chest.Position.Y
-
-										if isSameObj && !obj.Selectable {
-											return true
+
+							_, keyFound := d.Items.Find("Key", item.LocationInventory)
+
+							if chest.InteractType == object.InteractTypeLocked && !keyFound {
+								//Skip the chest if its locked and we don't have keys in our inventory
```
theksk23, line 64: kind of the same question I had for https://github.com/hectorgimenez/koolo/pull/148
You implement this here in a specific run directly: Any Idea about how we can get a global setting like maybe `Character.OpenChests`, and prevent putting this logic into every run we implement?

For example Tomb runs do want to open chests, and so run into this problem, too.
/whishlist feel free to close this comment without resolving, thanks!

nullptr-ref: The way I see it, due to the way its written, we will probably need a new method to replace InteractObject. If we create a method like InteractChest that has this check inside and it gets used instead of interactobject in all scripts then maybe, yea. But for now, this check is the only option we have. 

Another thought is to implement the openchest logic inside the clear room function with the setting you mentioned that will solve us having to manually implement openchests in every single script except the ones that only need to open chests where manual interactions will be required.

@hectorgimenez your thoughts on this?

hectorgimenez: I like the OpenChest action idea, we can embed this logic inside and just call InteractObject step if required or return nil otherwise. I don't know if it will be 1/1 replacement for all the usecases.

internal/supervisor.go
```diff
@@ -65,7 +67,16 @@ func (s *baseSupervisor) Stop() {
 	}
 
 	s.c.Injector.Unload()
+	process, err := os.FindProcess(int(s.c.Reader.Process.GetPID()))
```
hectorgimenez, line 70: I don't know if this is something we want to do? For example when I stop it, I like to go to the stash and see what I have there. But regardless if we want to start this discussion or not, I think we should do it in a different PR to not mix stuff

internal/run/lower_kurast_chests.go
nullptr-ref: Just to be on the same page, you're talking about implementing the openchests inside of the clear_area function, correct? 

We can also implement another function to open chests (like interactobject) to have the same check inside that can be used for scripts like lower kurast chest farm. Not sure if there are any other use-cases but might be nice to have, I suppose.

hectorgimenez: oh nono, I meant the second option, implement a new action like interactobject but doing the check inside

PR #152, opened 2024-04-19
Implement InteractChest

nullptr-ref:
Implemented a new method called InteractChest to be used when we want to open chests. The way it works is that it checks the chest type and if it's locked and you don't have a key in your inventory, it overrides the isCompletedFn to always return true, which itself causes the chest to get skipped.

Changed InteractObject to InteractChest in all scripts where applicable


internal/action/interaction.go
```diff
@@ -48,6 +49,41 @@ func (b *Builder) InteractNPCWithCheck(npc npc.ID, isCompletedFn func(d game.Dat
...
+				NewStepChain(func(d game.Data) []step.Step {
+					steps := []step.Step{step.InteractChest(o.Name, isCompletedFn), step.SyncStep(func(d game.Data) error {
+						event.Send(event.InteractedTo(event.Text(b.Supervisor, ""), int(name), event.InteractionTypeObject))
+						return nil
+					})}
+
+					return append(steps, additionalSteps...)
+				}),
+			}
+		}
```
hectorgimenez, line 83: What about this? So we only move/real attempt to open it if we have keys, otherwise we just return nil, doing nothing.
I think it's okay skip emitting the event, since it's a single interaction item, so companions will not be able to interact it.

```suggestion
		o, _ := d.Objects.FindOne(name)
		_, keyFound := d.Items.Find(item.Key, item.LocationInventory)

		if o.InteractType == object.InteractTypeLocked && keyFound {
			return []Action{
				b.MoveToCoords(o.Position, step.StopAtDistance(7)),
				NewStepChain(func(d game.Data) []step.Step {
					steps := []step.Step{step.InteractObject(o.Name, isCompletedFn), step.SyncStep(func(d game.Data) error {
						event.Send(event.InteractedTo(event.Text(b.Supervisor, ""), int(name), event.InteractionTypeObject))
						return nil
					})}

					return append(steps, additionalSteps...)
				}),
			}
		}

		return nil
```

internal/action/step/interact_object.go
```diff
@@ -29,6 +30,14 @@ func InteractObject(name object.Name, isCompleted func(game.Data) bool) *Interac
 	}
 }
 
+func InteractChest(name object.Name, isCompleted func(game.Data) bool) *InteractObjectStep {
```
hectorgimenez, line 33: I think we can use InteractObject, since it's the same?

internal/action/interaction.go
nullptr-ref: Thanks for the suggestion. I already tried to just return nil which backfired hard. For some reason, when you just return nil, it screws up the other chain/steps, leading to weird behavior. 

I tested specifically with the lower_kurast_chestrun.go script and that just led to error after error, skipping normal chests and other weird stuff. With this, it theoretically should just skip interacting with the chest no matter how many times its detected but I could be wrong ... 

I'll test my implementation again today (without keys in inventory) since I already have a compiled version of Koolo with the changes and see if it indeed gets stuck on locked chests (since I'll be running ancient tunnels & stony tomb)

btw @hectorgimenez I'll mail you later (on your publically shown email) with something I want to discuss with you to optimize the commits & contribution to the repo

PR #155, opened 2024-04-20
Added focusonelite & openchests option to more maps

nullptr-ref:
Added focusOnElitePacks & openChests configurations for the following maps:

- Stony Tomb
- Ancient Tunnels
- Pit (added the filter & openChests option to the clearArea call on lvl2)

Also updated the default template accordingly.


hectorgimenez:
Thanks!

PR #159, opened 2024-04-21
Check for max quantity during pickup as well

13413j1j13j5315n13:


internal/action/item.go
```diff
@@ -0,0 +1,63 @@
...
+
+	b.Logger.Debug(fmt.Sprintf("For item %s found %d max quantity from pickit rule, number of items in the stash tabs %d", i.Name, maxQuantity, matchedItemsInStash))
+
+	return matchedItemsInStash >= maxQuantity
+}
+
+func (b *Builder) allStashItems(d game.Data) (S []data.Item) {
+	return slices.Concat(
+		d.Items.ByLocation(item.LocationStash),
+		d.Items.ByLocation(item.LocationVendor),       // When stash is open, this returns all items in the three shared stash tabs
```
hectorgimenez, line 58: @13413j1j13j5315n13 is it correct to call it during item pickup? Can we get stash contents if the stash is not open?

PR #188, opened 2024-05-01
Added support for custom command line arguments

nullptr-ref:


jang9824:
d2R.exe -mod skinName 

-mod skinName 

Mode is not applied
How do I create a command line for mod application?

jang9824:
command line arguments

-Mode SkinName

Mode not applied

What do I need to do?

PR #194, opened 2024-05-02
Do not merge? Revert "Implement chickening on specific monsters"

13413j1j13j5315n13:
Reverts hectorgimenez/koolo#148


13413j1j13j5315n13:
Seems like https://github.com/hectorgimenez/koolo/pull/195 fixes chickening. So if that works properly after testing this PR needs to be declined.

PR #196, opened 2024-05-03
improve nova sorc's attack process

xg6636:
add 1 StaticField before other attacks,


nullptr-ref:
Casting static on every attack rotation or different mob will slow the attack process a lot. We need proper logic to check the monsters around and their current health to determine if a static should be cast or not

xg6636:
good point.

PR #210, opened 2024-05-09
Cubing recipes

nullptr-ref:
Added cubing_recipes functionality - Cube flawlesses to perfects, cube essenses into token
Added cubing recipes as configurable option in settings.
Updated code to switch to correct stash page when trying to interact with the cube


internal/action/stash.go
```diff
@@ -212,3 +212,16 @@ func (b *Builder) switchTab(tab int) {
 	b.HID.Click(game.LeftButton, x, y)
 	helper.Sleep(500)
 }
+
+func (b *Builder) SwitchTab(tab int) *Chain {
```
hectorgimenez, line 216:
```suggestion
func (b *Builder) SwitchStashTab(tab int) *Chain {
```

Builder struct is shared across multiple parts of the application, better to be specific.

internal/server/http_server.go
```diff
@@ -312,6 +312,11 @@ func (s *HttpServer) characterSettings(w http.ResponseWriter, r *http.Request) {
 		// Gambling
 		cfg.Gambling.Enabled = r.Form.Has("gamblingEnabled")
 
+		// Cube Recipes
+		cfg.EnableCubeRecipes = r.Form.Has("enableCubeRecipes")
+
+		// Companion
```
hectorgimenez, line 318: We don't need this?

nullptr-ref: Some people might not want to enable the cube recipes to keep their flawless gems for other purposes. Additionally, we might add other recipes in the future that people don't want to waste their resources on.

divineblade7: That was my thought as well, such as crafting recipes or cubing runes eventually. Having option to disable would be good.

PR #284, opened 2024-06-04
Reduced characters from doing a "leeroy jenkins" in chaos sanctuary

SoundsLegit:
Added a `MoveToCoordsWithMinDistance` function that allows to move to coordinates but not force continue if the character is within a certain minimum distance (which I set to 20). This allows the character to stop and fight the elite packs in chaos sanctuary instead of doing a leeroy jenkins into them.


SoundsLegit:
This branch needs more work, currently the character does no longer stupidly run into groups, but it sometimes fails to open seals.

PR #289, opened 2024-06-07
🎈 perf: add EnergyShield for SorcLevelingLightning if available.

xg6636:
in current version there is no **EnergyShield** for the lightning leveling char, but **EnergyShield** is crucial for this build, so i made a patch and wish that could help.


SoundsLegit:
added this fix and more to buffing for characters in the trapskin pull request.

xg6636:
> added this fix and more to buffing for characters in the trapskin pull request.

good job.

PR #291, opened 2024-06-07
Foh holybolt

SoundsLegit:
Added usage of holy bolt in between (approximate) cooldowns of fist of the heavens when fighting a demon-type boss (Mephisto, Andariel, Duriel, Diablo, Baal)

### DIFF
--- a/internal/character/foh.go
+++ b/internal/character/foh.go
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	fohMaxAttacksLoop = 10
+	fohMaxAttacksLoop = 20
 	fohMinDistance    = 5
 	fohMaxDistance    = 20
 )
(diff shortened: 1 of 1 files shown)


AquaRush:
Hi, it would be best to modify the script to work only with all the super bosses except Andariel.
Andariel area is filled with unique monsters sometimes, where FoH works well on cleaning them, it creates a safer environment for HC players :)
Thanks for the contribution!

SoundsLegit:
> Hi, it would be best to modify the script to work only with all the super bosses except Andariel. Andariel area is filled with unique monsters sometimes, where FoH works well on cleaning them, it creates a safer environment for HC players :) Thanks for the contribution!

The idea is not to _just_ use holy bolt on demon bosses, but to switch to using holy bolt 3 times quickly while FoH is on cooldown and then switch back to FoH and use it again. The character will still use FoH on andariel. This is the best of both worlds.
Also, to deactivate this feature, it is enough to just **not have the holy bolt/Foh mouse 1 hotkeys binded**

AquaRush:
> > Hi, it would be best to modify the script to work only with all the super bosses except Andariel. Andariel area is filled with unique monsters sometimes, where FoH works well on cleaning them, it creates a safer environment for HC players :) Thanks for the contribution!
> 
> The idea is not to _just_ use holy bolt on demon bosses, but to switch to using holy bolt 3 times quickly while FoH is on cooldown and then switch back to FoH and use it again. The character will still use FoH on andariel. This is the best of both worlds. Also, to deactivate this feature, it is enough to just **not have the holy bolt/Foh mouse 1 hotkeys binded**

I see, that makes more sense then, thanks!

SoundsLegit:
With the latest commit I made it simpler.
Holy bolt can be set as a mouse button 2 key bind, and the bot only does 2 holy bolts before switching back to FoH.

SoundsLegit:
Ready for merge on last commit, fixed the bug of potentially not switching back to FoH for normal mobs.

TDLofCC:
Been running this change on my FoH pala since it was made, seems to run very well.

PR #293, opened 2024-06-08
Nova sorc improvements

nullptr-ref:
Add check if we should use static in main attack rotation
* It will occur if more than 50% of mobs within max attack distance +5 are above 60% hp.

### DIFF
--- a/internal/character/lightning_sorceress.go
+++ b/internal/character/lightning_sorceress.go
@@ -33,6 +33,7 @@ func (s LightningSorceress) KillMonsterSequence(
 	previousUnitID := 0
 
 	return action.NewStepChain(func(d game.Data) []step.Step {
+
 		id, found := monsterSelector(d)
 		if !found {
 			return []step.Step{}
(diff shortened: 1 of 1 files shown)


internal/character/lightning_sorceress.go
```diff
@@ -97,6 +104,36 @@ func (s LightningSorceress) BuffSkills(d game.Data) []skill.ID {
...
 
+func (s LightningSorceress) shouldCastStatic(d game.Data) bool {
+
+	// Iterate through all mobs within max range and collect them
+	mobs := make([]data.Monster, 0)
+
+	for _, m := range d.Monsters.Enemies() {
+		if pather.DistanceFromMe(d, m.Position) <= lightningSorceressMaxDistance+5 {
+			mobs = append(mobs, m)
+		} else {
```
hectorgimenez, line 115: I think you don't need this else condition

hectorgimenez:
wdyt about adding a backup limit? for example to set max 10 static casts or something, regardless if the condition is still there. Just to avoid casting static field forever over a lightning inmune mob, or a mob behind a wall or any weird situation

PR #309, opened 2024-06-22
Buff refresh time interval as a setting

SoundsLegit:


artosimonyan:
This will probably get modified in the new version to take the buff duration into account.

PR #358, opened 2024-07-13
Repair throwables

SoundsLegit:


artosimonyan:
Obsolete with #422. Port it over to the new version, please!

PR #359, opened 2024-07-15
Fix index out of range error in LineOfSight and ensureEnemyIsInRange

R1DEL:
#### Changes:
1. **Added boundary checks in `LineOfSight` function:** 
   - Ensured that indices `x0` and `y0` are within the valid range of the `d.CollisionGrid` array to prevent out-of-bounds errors.

2. **Adjusted index calculation in `ensureEnemyIsInRange` method:**
   - Modified the calculation of `moveTo` to ensure it does not exceed the bounds of the `path` array.

#### Testing:
- Ran multiple scenarios to ensure the changes fix the index out of range error without introducing new issues.
- Verified that the bot performs as expected in all tested scenarios.

### DIFF
--- a/internal/action/step/attack.go
+++ b/internal/action/step/attack.go
@@ -160,7 +160,7 @@ func (p *AttackStep) ensureEnemyIsInRange(container container.Container, monster
 
 	path, distance, found := container.PathFinder.GetPath(d, monster.Position)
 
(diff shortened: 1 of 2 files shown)


artosimonyan:
Thanks for the contribution, will wait for @hectorgimenez to double-check :)

PR #366, opened 2024-07-17
Added crash detection

artosimonyan:


internal/game/crash_detector.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+}
+
+func (cd *CrashDetector) Stop() {
+	cd.logger.Info("Stopping Crash Detector", slog.Int("PID", int(cd.pid)), slog.String("Supervisor", cd.supervisor))
+	close(cd.stopChan)
+}
+
+func (cd *CrashDetector) isProcessRunning() bool {
+	_, err := process.NewProcess(cd.pid)
```
hectorgimenez, line 59: what if we use go windows library so we don't need the new extra dependency just for this?

Something like:

```
	handle, err := windows.OpenProcess(windows.PROCESS_QUERY_LIMITED_INFORMATION, false, pid)
	if err != nil {
		return false
	}
	defer windows.CloseHandle(handle)

	return true
```

PR #373, opened 2024-07-23
Discord webhook support for posting found items

klb92:
Support for simple discord webhook for posting found items to a channel, with an option to filter the items to post

image
image


artosimonyan:
Really nice addition, but we already have a discord implementation, although it's not this easy to set up. We'll discuss it internally if we want to have both implementations, just one and which one. 

Thank you for the contribution <3

PR #374, opened 2024-07-24
Update council.go

TDLofCC:
Changed tun to move to specific coordinates walking towards the council temple. 

Using variable coordinates because Travincal map seems to have various starting coordinates.

Clearing in front of temple, ignoring water snakes to prevent from getting stuck and clearing temple only targeting councilmembers. It no longer uses the function char.killcouncil(). So you could technically remove it from the character configs.

Tested with pala and sorc, but was mainly made for character without walking. Works very well with WW barb ;-)

### DIFF
--- a/internal/run/council.go
+++ b/internal/run/council.go
@@ -1,8 +1,11 @@
 package run
 
 import (
+	"slices"
+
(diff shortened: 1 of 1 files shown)


artosimonyan:
Obsolete with #422

PR #375, opened 2024-07-24
Update andariel.go

TDLofCC:
Before killing Andariel it first clears the room in front of her before attacking Andariel.

Much safer approach, especially early game.


nej-tak:
Been running this 100+ on nightmare using a blizzsorc, seems like a big improvement 🙌

PR #376, opened 2024-07-24
Update mephisto.go

TDLofCC:
Made Mephisto run a bit safer, especially nice for early game runs, instead of directly attacking bot moves to attack position, waits for Mephisto to move then moves to a safer attackposition and starts attacking him.

Doesn't always work, Mephisto movement seems to be random, but in test runs it really helped with getting the chicken count down.

After killing Mephisto bot goes through red portal to A4 for faster next game town movement.

### DIFF
--- a/internal/run/mephisto.go
+++ b/internal/run/mephisto.go
@@ -44,7 +44,7 @@ func (m Mephisto) BuildActions() []action.Action {
 		m.builder.Wait(time.Second * 1),                     // wait 1 second for Mephisto to move
 		m.builder.MoveToCoords(mephistoSaferAttackPosition), // move to safer attack position
 		m.char.KillMephisto(),                               // attack mephisto
-		m.builder.ItemPickup(false, 20),                     // making sure we pick up stuff before moving to A4 through red portal for faster next game town movement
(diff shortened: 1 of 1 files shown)


nej-tak:
I tried this 10 times on nightmare using a blizz sorc (bad gear so slow killtime).

Somehow the bot exits the area via the red portal **before** Meph dies, which I didnt even know was possible.
It seems to do it after navigating to the safespot.

(If I pause the bot after it enters act4 and then manually go back to meph again, in the same game, he is still alive)

I'll try to repro again tomorrow and catch a video, maybe sleep deprivation is getting to me

Millionaar:
I presume it will be the slow kill time. It happened when I was undergeared also, on nova sorc.
It will be triggering a time or attack loop limit, so the bot assumes the monster is unreachable etc.

Not sure on the fix/workaround, but take a look at lightning_sorceress.go where it casts static first on particular bosses. Some of that additional code might be a good starting point.
You should be able to add a specific override for Mephisto where it attacks more loops or something along those lines.

Sorry I'm not a developer but hope that points you in the right direction :)

G2G2G2G:
If you get ~10 levels into static you can hit him from far away

artosimonyan:
Obsolete with #422

PR #377, opened 2024-07-24
Update diablo.go

TDLofCC:
Tried to make Chaos Sanctuary runs a bit safer when popping the seals. It moves to specific safe coordinates after popping a seal to spawn the elite, based on the location of the seal coordinates.

Added extra forced buff times to make sure we always have BO.

Sometimes the bot wouldn't go back to the star to kill Diablo, added 3 extra coordinates to use to walk back to the star.

Changed the entranceToStar coordinates to more closely resemble a manual clear of monsters from entrance to star and making it a little safer.

### DIFF
--- a/internal/run/diablo.go
+++ b/internal/run/diablo.go
@@ -23,22 +23,24 @@ var diabloSpawnPosition = data.Position{
 	Y: 5286,
 }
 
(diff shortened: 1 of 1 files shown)


artosimonyan:
Closing in favor of #383

PR #381, opened 2024-07-28
Fix mosaic spelling

ElectricDosh:
Correct character profile spelling from "mosiac" to "mosaic".


ElectricDosh:
Note: if running a mosaic profile, it will be necessary to re-select the class in the character settings, as it will default back to blizzard sorc.

PR #401, opened 2024-08-11
Restructured conditionals

Farmith:
* Restructuring in an attempt to fix problem with constantly trying to repair 100% durability items

### DIFF
--- a/internal/action/repair.go
+++ b/internal/action/repair.go
@@ -32,8 +32,7 @@ func (b *Builder) Repair() *Chain {
 
 			// Restructured conditionals for when to attempt repair
 			if 	(maxDurabilityFound && !found) ||
-				(durabilityPercent != -1 && found && durabilityPercent <= 20) ||
-				(found && durability.Value <= 5) {
+				(durabilityPercent != -1 && found && durabilityPercent <= 20) {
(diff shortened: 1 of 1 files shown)


Farmith:
Added some restructuring as low level characters in particular constantly try to repair 100% durability items (such as self repairing ones, or low durability ones), this seems to alleviate it in my case

internal/action/repair.go
```diff
@@ -30,7 +30,9 @@ func (b *Builder) Repair() *Chain {
 				durabilityPercent = int((float64(durability.Value) / float64(maxDurability.Value)) * 100)
 			}
 
-			if maxDurabilityFound && !found || durabilityPercent != -1 && found && durabilityPercent <= 20 || found && durability.Value <= 5 {
+			// Restructured conditionals for when to attempt repair
+			if 	(maxDurabilityFound && !found) ||
+				(durabilityPercent != -1 && found && durabilityPercent <= 20) {
```
artosimonyan, line 35: We do need the value as a failsafe. Probably decrease it to 3, to account for low level items.

Farmith: Added back a failsafe

PR #405, opened 2024-08-12
A2/A3 leveling fixes

Farmith:
Small fix that makes so that A2 leveling works better (now doesn't crash when trying to leave town)


artosimonyan:
Obsolete with #422

PR #406, opened 2024-08-12
Add grid view, drop filtering, persistent drops, auto-start

ElectricDosh:
Changelog:

Updated Dashboard:
New grid view option for the dashboard, with toggles to switch between list and grid layouts.
View preferences are saved between sessions.
Added ability to expand all character cards simultaneously.
Added Total Drops counter, which leads to the combined drops page for All Characters.
New auto-start feature for sequentially starting multiple characters.

Enhanced drops page:
New filtering options for easier item searching.
View drops from all characters in a single, consolidated list.

General enhancements:
Drop logs now persist through bot crashes or manual stops, maintaining records until Koolo is closed.


artosimonyan:
Auto start was implemented via Scheduler. Drops interface will be reworked in the new version at some point. 
Thanks for the contribution though.

PR #407, opened 2024-08-12
fix issue where only one object with a given name would be added

nabwerks:
*log how many fires were found in lower kurast run


nabwerks:
LK runs were only able to find one chest when two were in the map. This is because we only add one object with a given name, which is fixed by this PR.

artosimonyan:
Thank you for the contribution. 

This will be implemented in #422

PR #413, opened 2024-08-14
Fix caves with doors

Farmith:
* This fix solves a pathing issue with "caves with doors" as in A2, with this in place the toons can safely navigate Halls of the Dead Level 1,2,3 and Maggot Lair 1,2,3


artosimonyan:
Obsolete with #422

PR #415, opened 2024-08-15
Fix repair loop, throwables

ElectricDosh:
Replaces the check in the Repair() function with RepairRequired() to ensure we always get the same result when we check in town as we do when we check mid-run. This should prevent the endless tp loop bug.

Also replaces the RepairRequired() function with the one from SoundsLegit's repair_throwables branch.


RichRadd (approved):

artosimonyan:
Obsolete with #422

PR #417, opened 2024-08-15
Initial work for shrines

Farmith:
* Initial work for shrines, added the base class/shrines into game


artosimonyan:
Obsolete with #422. Please port it over.

PR #418, opened 2024-08-15
Prevent Assassins buying keys

ElectricDosh:
Update the BuyConsumables() function in shop_manager.go so we no longer buy keys if character class = mosaic, since assassins don't need keys to unlock locked chests.


SoundsLegit:
If you have no keys it shouldn't buy keys

ElectricDosh:
> If you have no keys it shouldn't buy keys

It does currently buy keys for assassin regardless of whether or not we already have any, and I couldn't find any check in the existing code that would stop it doing so.

artosimonyan:
Thanks for the contribution. 

Will be implemented with #491 #422

PR #419, opened 2024-08-15
Update interact_object.go

ElectricDosh:
- Added timeout functionality to InteractObjectStep in interact_object.go
- Implemented automatic movement to a random point in a 10x10 grid centred on the object after 8 seconds of failed interaction attempts
- Limited to 5 movement attempts
- Added completion check at the start of each interaction attempt
- Preserved original object interaction logic

### DIFF
--- a/internal/action/step/interact_object.go
+++ b/internal/action/step/interact_object.go
@@ -1,125 +1,174 @@
 package step
 
 import (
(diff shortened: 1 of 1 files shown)


artosimonyan:
Obsolete with #422

PR #432, opened 2024-08-17
visit chests in efficient order in LK chest run

nabwerks:


artosimonyan:
Thank you for the contribution! 

This will be added with #422 in the new way we're handling the actions.

PR #434, opened 2024-08-17
Improvements to setting up Leveling characters

Farmith:
* Added search field for disabled runs to find what you are looking for faster
* Added "clear all" button to enabled runs to clear the enabled runs faster
* Added popup mentioning that the leveling character profiles/classes require enabling a leveling run profile, and asks user to either confirm or deny this automated action which will clear all enabled run profiles and auto enable the leveling run profile


Farmith:
The force push was just to clean this PR, accidentally got some commits in that weren't part of this PR

PR #435, opened 2024-08-18
Add supervisor name to window title

Farmith:
* Added functionality that adds the supervisor name to the game client title to easier find which of all windows is having an issue and solve it


artosimonyan:
Implemented with #490

PR #448, opened 2024-08-19
Added some fixes for coldplains leveling

Farmith:
* useWaypoint instead of MoveTo to remove a bug where toon is lvl 3+ but doesn't know how to get to CP


artosimonyan:
This is way outdated now :D

PR #467, opened 2024-09-04
Update diablo.go

elobo91:
Fixed issue with Diablo Run where walkable character would not reach diablo after interacting with last seal (save and exit game).  Now compatible with both teleporting/walkable characters.

Update 2: Now skip (abort run) if kill diablo is unchecked after last seal
update 3 : Removed  unused MoveToDiabloSpawn()


elobo91:
You are right i carried an unused function . I will correct this. However movement being call at the end of infector part is intended .  In v2 refactored version it could be outside of it but right now it wont work if i let it at the same place it was.

artosimonyan:
Obsolete with #422

PR #474, opened 2024-09-11
Fix item pickup

artosimonyan:


internal/action/step/pickup_item.go
```diff
@@ -25,19 +25,24 @@ func PickupItem(it data.Item) error {
 	mouseOverAttempts := 0
 	currentMouseCoords := data.Position{}
 	lastRun := time.Now()
+	itemToPickup := it
 
 	for {
 		// Pause the execution if the priority is not the same as the execution priority
 		ctx.PauseIfNotPriority()
+		ctx.RefreshGameData()
```
hectorgimenez, line 33: this is the key part

internal/action/step/pickup_item.go
```diff
@@ -54,10 +59,16 @@ func PickupItem(it data.Item) error {
...
 		lastRun = time.Now()
-		objectX := it.Position.X
-		objectY := it.Position.Y
+		objectX := it.Position.X - 1
+		objectY := it.Position.Y - 1
+
 		mX, mY := ui.GameCoordsToScreenCords(objectX, objectY)
 
+		// Move the mouse to the coords
+		ctx.HID.MovePointer(mX, mY)
```
hectorgimenez, line 68: we should refresh game data again after this movement if we want to be able to detect the hovered item during this same loop iteration, otherwise the if it.IsHovered will be false even if the mouse is on top of the item (for this iteration)

PR #503, opened 2024-10-01
Fixes for legacy graphic and more

elobo91:
Fixes to ensure game is loaded before performing anything. We also want to make sure that legacygraphics are active before performing anything because that will crash anyone using mods. It also fix issue where it would fail identifying an item and have it on cursor spiraling forever.


elobo91:
all last 3 commits goes together thats why fail check. . sorry for not putting in the same

PR #504, opened 2024-10-01
UsePortalInTown()  pickupitems

elobo91:
If run finished and bot trigger a context to do towning actions( like full bag) when returning to area using tp it wont pickup remaining items.It would either save/exit game or switch to next run on the list.    --Added a pickupitem there when reentering portal.


elobo91:
all last 3 commits goes together thats why fail check. . sorry for not putting in the same

PR #506, opened 2024-10-03
Berserk_barb arrives to v2 (travincal barb for now)

elobo91:
This is a fully working travincal berserk barbarian profile.
Its using FindItem on 100% of the corpses. 

Made some exception to barbarian attacks as he is doing melee  and his skill (berserk) has no cooldown.   Normal primaryattack was  making the attacks 2x slower if not more.

It checks if it accidently clicked on durance lvl 1 stairs  then come back to travincal and resume its task. (there could be optimization here it takes up to 15 seconds but works)

had to lock the context to pickup items after killing a monster(only for council) . we must kill all 12 council members before looting/horking . thats why the exception.

- General koolo Fixes
added a  WaitForGameToLoad()  in bot.go  that make sure game is fully loaded  and apply legacygraphics ( if classic mode enabled)  before performing anything.   That would cause  instant crash for anyone using mods.   that was also causing a problem when identifying items at the same time as toggling legacy mode.  The change from HD to legacy was making the bot misclick and put item on cursor that would spiral forever.    

Added a pickup_item  after returning from  town .   If run is done and character  had full inventory then it triggered the return to town context.  When it would come back the context was done (the run) so it wouldnt pickup remaining items  then it would either save/exit game or  continue with next run on the list.    item_pickup added in UsePortalInTown() to prevent this.

This is not a finished product / i have so much more to add but that would become very long to review so lets have a first version.     ETA:  very solid. can run without interuptions having teleport from enigma .  Theres minor issue with walking character  the  context  "WaitForAllMembersWhenLeveling"  gets triggered even tho its not a lvling profile .  it happens occasionally but it doesnt with teleport .  

elb

### DIFF
--- a/internal/action/tp_actions.go
+++ b/internal/action/tp_actions.go
@@ -59,8 +59,7 @@ func UsePortalFrom(owner string) error {
 
 			if err == nil {
 				// Perform actions after re-entering the game area
-				ItemPickup(30)   // Pick up items in the immediate area
-				BuffIfRequired() // Reapply buffs if needed
+
 			}
 
 			return err
--- a/internal/action/tp_actions.go
+++ b/internal/action/tp_actions.go
@@ -49,16 +49,23 @@ func UsePortalFrom(owner string) error {
 
 	for _, obj := range ctx.Data.Objects {
 		if obj.IsPortal() && obj.Owner == owner {
-			return InteractObjectByID(obj.ID, func() bool {
+			err := InteractObjectByID(obj.ID, func() bool {
 				if !ctx.Data.PlayerUnit.Area.IsTown() {
 					utils.Sleep(500)
 					return true
 				}
-
 				return false
 			})
+
+			if err == nil {
+				// Perform actions after re-entering the game area
+				ItemPickup(30)   // Pick up items in the immediate area
+				BuffIfRequired() // Reapply buffs if needed
+			}
+
+			return err
 		}
 	}
 
-	return nil
+	return errors.New("portal not found")
 }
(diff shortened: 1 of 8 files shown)


internal/bot/bot.go
```diff
@@ -108,9 +117,16 @@ func (b *Bot) Run(ctx context.Context, firstRun bool, runs []run.Run) error {
 					continue
 				}
 
+				if b.ctx.CharacterCfg.ClassicMode && !b.ctx.Data.LegacyGraphics {
```
hectorgimenez, line 120: Maybe we can make this check inside the SwitchToLegacyMode function + the ctx.RefreshGameData()? wdyt? so we can also get rid of the ctx.RefreshGameData() on line 47

internal/action/tp_actions.go
```diff
@@ -49,16 +71,22 @@ func UsePortalFrom(owner string) error {
...
 				if !ctx.Data.PlayerUnit.Area.IsTown() {
 					utils.Sleep(500)
 					return true
 				}
-
 				return false
 			})
+
+			if err == nil {
+				// Perform actions after re-entering the game area
```
hectorgimenez, line 83: you wanted to put something here?

elobo91: I got confused with the
func UsePortalFrom(owner string)  and UsePortalInTown().  So i carried that change with the pr. I kept it like this if eventually we want to add stuff there . Is not necessary however

internal/bot/bot.go
elobo91: The extra ctx.RefreshGameData after switchtolegacymode is to make sure the context doesn’t switch back to HD having unrefreshed data. That happened when testing so i added it there. This doesnt affect performance . But maybe theres place for optimization

PR #513, opened 2024-10-15
Run statistics are back, improved nova/foh, bosses act correctly detected, pause button fixed, moveto improved, area correction +++

elobo91:


elobo91:
Some info on Areacorrection :  Travincal and Pit are both AreaAware ( pit as an exemple on how to use it) .    Travincal will return to adjacent level (travincal) if accidently entered durance  1 . if fails moving back then it will try from town  making a portal,  using waypoint then returning to last recorded movetocoords position ( wich is council position).   The start from town in travincal works with any character except for berserk_barb because of his iskillingcouncil .   By returning to town  berserk_barb returns that all council are dead so it proceed with horking phase.  i didnt have time to look into it but thats only him.  work flawless if using adjacentlevel method for him .    So basicly  you can pause bot with a character while killing council,  waypoint anywhere in game then resume bot.  it will return exactly where it was and resume killing.      I did an exemple for pit.    It keep a trace of visited areas  and also its progress( like clearlevel1  done,  clearlevel 2 done) .   so if you pause bot  while its clearing pit lvl 1  and  waypoint anywhere in game.  it will portal town,  use waypoint then move back sequentially through visitedareas  then resume at the lastaction area( run progress)  like pit lvl 1.    Exemple  if pit lvl 1 was done   then you waypoint to another area and resume  it will skip the clearing of lvl 1 and move directly to pit lvl 2 and resume clearing there.

elobo91:
Sometimes bot iddle  in a zone at any moment.  This is caused by   shouldbepickedup, pickupitems, getitemstopickup.    It idle for up to 15 seconds .  When pressing E  bot resume  so  I tried to add 

// Check if character is standing idle outside town
		if ctx.Data.PlayerUnit.Mode == mode.StandingOutsideTown {
			idleCounter++
			if idleCounter >= maxIdleCount {
				ctx.Logger.Debug("Character idle for too long, forcing movement")
				ctx.HID.PressKeyBinding(ctx.Data.KeyBindings.ForceMove)
				idleCounter = 0
				continue
			}
		} else {
			idleCounter = 0
		}

but then i realised it wasnt only happening in  pickupitems  but at 3 different places.   I applied the anti-idle to all 3 but that made the pickup_items delayed.    Have to find a better way to call the anti-idle  . ctx.Data.PlayerUnit.Mode == **mode.StandingOutsideTown  is the key.**

elobo91:
Berserker Specific options.    Now  FindItemSwitch option only appears when berserk_barb is the current profile.   Also added an option to   skippotionspickupduringtravincal   . is waste of time because it will refill next game anyway.     This can be reused for any class specific options ( only show up when that profile is selected).   Nova is an exemple.

elobo91:
move.go   had a slight modification  for  movetoarea  that involve interacting with entrance.   it would fail  1/2  time.

	if lvl.IsEntrance {
		maxAttempts := 3
		for attempt := 0; attempt < maxAttempts; attempt++ {
			// Add a short delay before each attempt
			utils.Sleep(200)

			err := step.InteractEntrance(dst)
			if err == nil {
				// Successful interaction, exit the loop
				break
			}

			if attempt == maxAttempts-1 {
				// If this was the last attempt, return the error
				return fmt.Errorf("failed to interact with entrance after %d attempts: %w", maxAttempts, err)
			}

		}

		// Add a short delay after successful interaction
		utils.Sleep(200)
	}

now have a retry mechanism.   solving the issue

elobo91:
For run statistics i tried to keep like it was before as much as i could but that wasnt working at all.   run statistics were not updating.

now we have 
func (s *SinglePlayerSupervisor) updateRunStats(runName string, startTime, endTime time.Time, reason event.FinishReason) {
	event.Send(event.RunStatsUpdated(event.Text(s.name, fmt.Sprintf("Updated stats for run: %s", runName)), runName, startTime, endTime, reason))
}
calling a new event  in   events.go .  

i added  to bot.go in Low priority loop  to update run statistics otherwize no statistics other then gamefinished.

var runFinishReason event.FinishReason
			if err != nil {
				switch {
				case errors.Is(err, health.ErrChicken):
					runFinishReason = event.FinishedChicken
				case errors.Is(err, health.ErrMercChicken):
					runFinishReason = event.FinishedMercChicken
				case errors.Is(err, health.ErrDied):
					runFinishReason = event.FinishedDied
				default:
					runFinishReason = event.FinishedError
				}
			} else {
				runFinishReason = event.FinishedOK
			}

			event.Send(event.RunFinished(event.Text(b.ctx.Name, fmt.Sprintf("Finished run: %s", r.Name())), r.Name(), runFinishReason))

internal/context/context.go
```diff
@@ -134,3 +151,9 @@ func (ctx *Context) WaitForGameToLoad() {
 	// Add a small buffer to ensure everything is fully loaded
 	time.Sleep(300 * time.Millisecond)
 }
+
+// Syncing player area with gamedata. fix the pathing error after entering a new area.
+func (ctx *Context) UpdateArea(newArea area.ID) {
+	ctx.Data.PlayerUnit.Area = newArea
```
hectorgimenez, line 157: This is not a good idea, it will be concurrently overwritten by the routine reading the game memory

internal/run/cows.go
```diff
@@ -79,32 +83,100 @@ func (a Cows) getWirtsLeg() error {
...
+	a.ctx.Logger.Info("Moving to Wirt's corpse",
+		slog.Any("path_length", len(path)),
+		slog.Int("distance", distance))
+
+	return action.MoveTo(func() (data.Position, bool) {
+		return wirtLocation, true
+	})
+}
+
+func (a Cows) findCairnStones() error {
```
hectorgimenez, line 150: this is not being used?

PR #531, opened 2024-11-20
Added Normal Gem Recipes

IronMaXxX:
Updated code to enable cubing normal gems to flawless gems

### DIFF
--- a/internal/action/cube_recipes.go
+++ b/internal/action/cube_recipes.go
@@ -20,7 +20,37 @@ type CubeRecipe struct {
 var (
 	Recipes = []CubeRecipe{
 
-		// Perfects
+		// Flawless Gems
+		{
+			Name:  "Amethyst",
+			Items: []string{"Amethyst", "Amethyst", "Amethyst"},
+		},
+		{
+			Name:  "Diamond",
+			Items: []string{"Diamond", "Diamond", "Diamond"},
+		},
+		{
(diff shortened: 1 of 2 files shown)


internal/action/cube_recipes.go
```diff
@@ -20,7 +20,37 @@ type CubeRecipe struct {
 var (
 	Recipes = []CubeRecipe{
 
-		// Perfects
+		// Flawless Gems
+		{
+			Name:  "Amethyst",
```
artosimonyan, line 25: The recipe name here needs to match with the recipe.go in the config. (this is for all newly added recipes). Look how the other recipes are added.

internal/config/recipes.go
```diff
@@ -2,6 +2,13 @@ package config
 
 var AvailableRecipes = []string{
 	// List out all the recipe names from cube_recipes.go
+	"Flawless Amethyst",
```
artosimonyan, line 5: These recipes aren't defined.

internal/action/cube_recipes.go
IronMaXxX: Thanks for the hint, addes Flawless to the recipes

internal/config/recipes.go
IronMaXxX: fixed recipe names in cube_recipes.go, now they are defined.

PR #532, opened 2024-11-20
Gambling and Stash Error Rework

Neo-Silver:
gambling.go
- added an intrinsic timer to prevent bot stuck ingame
- added a new function to recognize all items at the same time 
-- checks all items stated by the user simultanously 
-- creates a map of bought items
-- counts to 10 for each stated
-- once one reaches 10, it neglects it
-- once all reach 10 the map is reset

cube_recipes.go
- fixed the gamble() call to match the new gambling function

town.go
- fixed the gamble() call to match the new gambling function
- added an item identification call at the start of the prerun to not stash unwanted items

### DIFF
--- a/internal/action/cube_recipes.go
+++ b/internal/action/cube_recipes.go
@@ -2,6 +2,7 @@ package action
 
 import (
 	"slices"
+	"time"
 
 	"github.com/hectorgimenez/d2go/pkg/data"
 	"github.com/hectorgimenez/d2go/pkg/data/item"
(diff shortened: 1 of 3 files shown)


internal/action/cube_recipes.go
```diff
@@ -381,7 +382,7 @@ func CubeRecipes() error {
 
 				// TODO: Check if we have the items in our storage and if not, purchase them, else take the item from the storage
 				if recipe.PurchaseRequired {
-					err := GambleSingleItem(recipe.PurchaseItems, item.QualityMagic)
+					err := GambleSingleItem(recipe.PurchaseItems, item.QualityMagic, time.Now())
```
artosimonyan, line 385: Why is the current time an argument, forcing the package include in multiple files? Just get the current time when the function is called.

Neo-Silver: You are absolutly right.
I changed it to be an intrinsic timer only in the gambling.go
I changed back the calls in town and cube_recipes.

PR #544, opened 2024-11-29
fix mkdir

baiachen:
use ```os.MkdirAll()```


artosimonyan:
What does this address and why should we switch?

PR #549, opened 2024-12-06
Leveling improvements

CarlPoppa1:
Doesn't need TP tome bound - it gets bound when one gets bought
Added checks for TP tome and charges before trying to open portal
Fixed skill / stat allocation and key binding
Changed some stat allocation logic so we have enough str to equip armour
Changed some skill allocation logic
Changed act 1 levelling runs to just run den / stony field / countess until we can teleport
Added custom countess run to use clearthroughpath to get exp in tower lvl 1 -5


CarlPoppa1:
PR id a mess - gonna re-create when I've made more progress

PR #555, opened 2024-12-08
Tristam no portal on gold piles

elobo91:


elobo91:
not needed anymore.  portal spiral has been buffed

PR #558, opened 2024-12-10
Auto socket when levelling

CarlPoppa1:
Add ability to automatically add items (runes) to sockets, based on cube recipe code


artosimonyan:
Interesting implementation. Are you willing to expand upon this further and maybe rename it for clarity? 

Additional notes:
- Always put the main function that starts the loop at the top of the script so it can be located easily.

CarlPoppa1:
Yeah, how do you think it could be expanded? More runewords? Extra functionality? Also happy to take recommendations for a better name. I'll move the main function to the top, thanks for your help

PR #559, opened 2024-12-11
Sorceress classes with Staticfield

DarkDBx:
Added Lightning Sorc, added StaticField to Blizzard and HydraOrb Sorc, adjusted Distance to Attack


artosimonyan:
No no no no, not merging this :D 

- One change per PR relating to the PR title
- New builds in separate PRs as well, please!
- If you plan to add static to sorceresses' builds other than nova, you need to take the range of the mob and static skill level into consideration as static doesn't have a large range with 1 base point in it.

PR #568, opened 2024-12-13
Fix missed buff ( mosaic and others)

elobo91:


elobo91:
Can be closed, another pr fixed it

PR #569, opened 2024-12-13
enable CainIdentify with CharacterCfg option

srliao:
Added option to enable the existing `CainIdentify` code from [here](https://github.com/hectorgimenez/koolo/commit/5cbde98842e12c1863df7b136aef980c534c3919). Tested on SP and it seems to work pretty well.


srliao:
Added option to UI and tested to make sure it saves to the config file

{58319C05-6801-46F5-81BD-29A00B48678E}

artosimonyan:
Hehe, forgot to enable this as I didn't test it extensively after adding it. Glad to hear it's working as expected.

Few notes though:
- Wording for the UI option is bad. Change it to Identify with Cain, omit the long explanation afterward.
- Move the UI element above the minimum gold option
- Most important: We need fallback logic in case identifying with Cain fails. For example if Cain wasn't found (ie wasn't rescued), it should fallback to normal way of identifying.

srliao:
> Hehe, forgot to enable this as I didn't test it extensively after adding it. Glad to hear it's working as expected.
> 
> Few notes though:
> 
>     * Wording for the UI option is bad. Change it to Identify with Cain, omit the long explanation afterward.
> 
>     * Move the UI element above the minimum gold option
> 
>     * Most important: We need fallback logic in case identifying with Cain fails. For example if Cain wasn't found (ie wasn't rescued), it should fallback to normal way of identifying.

Thanks for the comments. I'll get the first two points fixed. Regarding the fallback logic, we'll need this PR merged first unfortunately: https://github.com/hectorgimenez/d2go/pull/54 as currently the quest data is not present.

In addition, the `UpdateQuestLog` function is currently bugged because https://github.com/hectorgimenez/koolo/blob/72ea47af0c5836d7cfbfc42c36555b5c9f6ca042/internal/action/step/close_all_menus.go#L16 does not actually check if QuestMenu is open. Related discussion here: https://discord.com/channels/1163042152985661480/1316961789191389254/1317256132594765897. I'll submit a separate PR to fix this

artosimonyan:
Revert this: https://github.com/hectorgimenez/koolo/pull/569/commits/09ac51a12d25a7b7dd6bc09852f8b4eb3ccbe9d5. 

We don't need anything from d2go. If it doesn't find Cain in town, the Identify with Cain will return an error, in which case we can proceed and use old logic. 

Don't try to overly complicate stuff.

srliao:
> Revert this: [09ac51a](https://github.com/hectorgimenez/koolo/commit/09ac51a12d25a7b7dd6bc09852f8b4eb3ccbe9d5).
> 
> We don't need anything from d2go. If it doesn't find Cain in town, the Identify with Cain will return an error, in which case we can proceed and use old logic.
> 
> Don't try to overly complicate stuff.

Done

srliao:
> Maybe switch line 119 and 120 in character_settings.gohtml, so the checkbox for "Use teleport when available" is also in front of the text like all other checkboxes.
> 
> grafik
> 
> grafik

I assume that should be a separate PR since it's an unrelated change to Cain? I can make the change on this PR if the maintainer wants me to though

PR #574, opened 2024-12-14
added chainlightning sorc

DarkDBx:
uses staticfield and lightning at bosses and chainlightning in other cases.


DarkDBx:
if you need to check the light script, compare it to nova.
i did used this as template.

Antizerg:
any update on getting this added to the main branch?  chain lightning sorc is a fairly popular build that is missing from Koolo

PR #585, opened 2024-12-17
Prevent isrepairrequired to loop when changing area

elobo91:
Goroutine check in loop if we need to repair. If we returnintown then data needed to evaluate will be temporary inaccessible. Added a timer after it was triggered to allow bot enough time to load data when using tp


elobo91:
closing to clean up..

PR #593, opened 2024-12-20
CharCfg-CheckboxTeleport

IronMaXxX:
Changed checkbox and text for the teleport option in char config (cosmetic thing to flatter my eyes :))


artosimonyan:
@IronMaXxX It's already like that in latest main:
{B2A82271-E14E-4517-8559-AA9AF6FDD053}

PR #594, opened 2024-12-20
Option Hide Portraits(prevent missing merc items)

elobo91:
New Configurable Option in Character settings to hide mercenary/ other players portraits.   Prevent misclick and opening merc inventory.   Will happen at same time as togglelegacy.


elobo91:
Closing to clean up...

PR #595, opened 2024-12-20
Added option for stashing items before first run

Sunblood:


artosimonyan:
We'll be removing the first run thing so no need for this one, thank you for the contribution though.

PR #598, opened 2024-12-20
clear_level.go optimization (ClearRoom)

elobo91:


davidfvsilva:
I tested this and it seems to work.

elobo91:
> I tested this and it seems to work.

Yes it works.  reduce failed to clear room in pit run as exemple from 11-12 error to 1-2 ( no walkable point in room to tiny)  .  i will however have to rewrite a little to match new pathfinder

elobo91:
This pr is still nice but closed for rework when AquaRush is done with new pathfinder

PR #599, opened 2024-12-21
Interactions reworked, area correction, repair loop fix

elobo91:
- New **AdaptativeSpiral** using desc from object.txt

- Multiple Completion checks for different objects wich allow to remove EnsureAreaAsync()

- Area correction only enabled for Travincal to prevent misclick on stairs leading to dura 1

- Now using Portal.Dest (area) for red portals and tome portals  instead of hardcoded destination.

- Cow level has configurable option to kill rakanishu otherwize will skip and interact with portal directly. Spiral is good enough for that.

- Cow level:  wont make portal directly on gold piles anymore

- Correction for Harem/palace cellar/ etc where there are 2 entrances for same destination

**Changes in Town:**  

- Prevent  town actions if we aren't in town , bot would identify items if didnt make it to town.

**Fix repair bug :**
when bot detected broken item (inreturnrownroutine) it would do townactions then after going back in portal the check in main loop would detect broken items again because we didn’t repair items while in town creating infinite portal loop.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/bwmarrin/discordgo v0.28.1
 	github.com/go-telegram-bot-api/telegram-bot-api/v5 v5.5.1
 	github.com/gorilla/websocket v1.5.3
-	github.com/hectorgimenez/d2go v0.0.0-20241216152730-04f45d306978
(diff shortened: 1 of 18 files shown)


elobo91:
Closed for massive improvement

PR #612, opened 2025-01-02
add throttling to game data refresh

srliao:
Currently, refreshes are not concurrency safe because multiple goroutines can (and have been) making mutable changes to `ctx.Data`

This PR makes `RefreshGameData()` concurrency safe and additional adds throttling to `RefreshGameData()` calls, which should result in a small performance increase, particularly in runs/characters that make excessive refresh calls

Tested this for about 12 hours now with no issues noted

### DIFF
--- a/internal/context/context.go
+++ b/internal/context/context.go
@@ -53,6 +53,10 @@ type Context struct {
 	LastBuffAt        time.Time
 	ContextDebug      map[Priority]*Debug
 	CurrentGame       *CurrentGameHelper
+
+	// for game data refreshes
+	reqRefresh chan chan bool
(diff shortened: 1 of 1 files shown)


srliao:
thanks for the comment. will shelf for now. the code has more fundamental issues as there's not really a good place to exit the goroutine used to throttle requests. 

and yes i'm aware it won't have significant impact, as it doesn't stop the consumer from spamming RefreshGameData() when it's not needed (nor does it address any inefficiency in GetData in d2go). the primary goal was actually to make a start at fixing a lot of the race condition present in the code (multiple goroutine concurrently accessing and modifying same variable).

PR #622, opened 2025-01-07
Item Pickup revision

elobo91:
- Following corrections on ClearAreaAroundPosition wich now doesnt return until everything under giving radius is dead (item in this case).  It shouldnt pickup item anymore when surrounded by monsters.
-  Added a line of sight check inside pickup_item action  in case item dropped on the other side of a wall and we cant pickup from here even if we can hover it.   If ErrNoLOSToItem , it will try to move  further the item because if we simply move at item position it wont work. we still will have no line of sight.
- func moveBeyondItem    is similar to the one in EnsureEnemyIsInrange to overshoot distance,  might consider moving this to utils.
-  new ItemSpiral  that is more precise to items instead of using generic  Spiral  currently used by everything else .  Prevents misclick on UI bar( belt, unbinding skills by accident)  being less wide

Logic is :  after detecting item,  we clear around item with a radius of 4 .  If we havent moved away to much from item  (distance <=4) we can attempt to pickup item directly.  otherwize we try to move closer with a distancetofinishmovement of 3 (thats perfect melee and interaction range).   Now its passed to  Pickup_item step  wich verify if we have line of sight,  also double check distance and attempt to pickup item.  If we get  ErrNoLOSToItem  it will go back to action item_pickup  and moveBeyondItem  (item position +2 distance). it will attempt to  step  pickup_item again after.     If for some reason it failed to pickup item,  added a second  ClearAreaAroundPosition  with smaller radius (3)  just in case some monsters added/were undetected  before giving up.  

If all attempts failed.. blacklist =(

### DIFF
--- a/internal/action/item_pickup.go
+++ b/internal/action/item_pickup.go
@@ -51,7 +51,6 @@ func ItemPickup(maxDistance int) error {
 	const maxRetries = 3
 
 	for {
-		// Get all items that should be picked up
 		itemsToPickup := GetItemsToPickup(maxDistance)
 		if len(itemsToPickup) == 0 {
 			return nil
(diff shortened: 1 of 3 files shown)


elobo91:
resurrected spacing compared to legacy 
Q20trjy

PR #628, opened 2025-01-15
Additional vendor check for keys and scrolls

sorc653:
The vendor selected for each act doesn't always have keys or scrolls, so we have to go to another vendor to do so.

This PR adds an additional check for purchasing keys and scrolls to ensure we don't run out. It also adds an option to return to town if we run out of keys. And it will only pick up keys if that option is enabled, or they already had some keys (but less than 12) in their inventory. It will not pickup more to put us over one stack.

Specific scenarios this fixes:
1. Act 2 scrolls will run out and not be re-purchased
2. Act 3 keys will run out and not be re-purchased

# Interface change:
Screenshot 2025-02-03 at 2 12 35 PM

### DIFF
--- a/internal/town/shop_manager.go
+++ b/internal/town/shop_manager.go
@@ -13,6 +13,37 @@ import (
 	"github.com/hectorgimenez/koolo/internal/ui"
 )
 
+
+func BuyTPs() {
+	ctx := context.Get()
+
+	if _, found := ctx.Data.Inventory.Find(item.TomeOfTownPortal, item.LocationInventory); !found {
+		ctx.Logger.Info("TP Tome not found, buying one...")
+		if itm, itmFound := ctx.Data.Inventory.Find(item.TomeOfTownPortal, item.LocationVendor); itmFound {
+			BuyItem(itm, 1)
+		}
+	}
+	ctx.Logger.Debug("Filling TP Tome...")
+	if itm, found := ctx.Data.Inventory.Find(item.ScrollOfTownPortal, item.LocationVendor); found {
+		BuyFullStack(itm)
+	}
+}
+
+func BuyIDs() {
+	ctx := context.Get()
+
+	if _, found := ctx.Data.Inventory.Find(item.TomeOfIdentify, item.LocationInventory); !found {
+		ctx.Logger.Info("ID Tome not found, buying one...")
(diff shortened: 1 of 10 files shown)


internal/town/shop_manager.go
```diff
@@ -113,7 +126,7 @@ func ShouldBuyIDs() bool {
 func ShouldBuyKeys() (int, bool) {
 	keys, found := context.Get().Data.Inventory.Find(item.Key, item.LocationInventory)
 	if !found {
-		return 12, false
+		return 0, true
```
sorc653, line 129: If we don't find any keys, we should return `0, true`, correct? The method is `ShouldBuyKeys`. So if we have none, we should return true saying we should buy keys.

I'm not missing something obvious here am I?

elobo91:
Hello, could you add keys to item_pickup in shouldbepickedup . Keys drop often from chests and it prevent going to town for nothing while it can sustain with drops. You will need to check stack quantity on ground if it fits stack in inventory ( max 11) . Key drop can be 1-2-3-4 key stacks . We cant use .nip for that as it wouldnt respect max key count and make multiple stack. Pr look good otherwise thanks

sorc653:
> Hello, could you add keys to item_pickup in shouldbepickedup . Keys drop often from chests and it prevent going to town for nothing while it can sustain with drops. You will need to check stack quantity on ground if it fits stack in inventory ( max 12) . Key drop can be 1-2-3-4 key stacks . We cant use .nip for that as it wouldnt respect max key count and make multiple stack. Buying at vendor when 1 key left( tp to town to refill) For scrolls vendor is totally fine. Pr look good otherwise thanks

Can you merge this [PR](https://github.com/hectorgimenez/d2go/pull/73) so I can use this method?

sorc653:
> Hello, could you add keys to item_pickup in shouldbepickedup . Keys drop often from chests and it prevent going to town for nothing while it can sustain with drops. You will need to check stack quantity on ground if it fits stack in inventory ( max 12) . Key drop can be 1-2-3-4 key stacks . We cant use .nip for that as it wouldnt respect max key count and make multiple stack. Buying at vendor when 1 key left( tp to town to refill) For scrolls vendor is totally fine. Pr look good otherwise thanks

Change complete!

hectorgimenez (approved):

PR #630, opened 2025-01-15
Repair loop fix, add support for items with quantity

Sunblood:
Add evaluation rules for items with quantity (javelins)
Fix going to town for repair and not repairing by removing check in Repair() that was different from RepairRequired()

### DIFF
--- a/internal/action/repair.go
+++ b/internal/action/repair.go
@@ -20,102 +20,86 @@ func Repair() error {
 	ctx := context.Get()
 	ctx.SetLastAction("Repair")
 
-	for _, i := range ctx.Data.Inventory.ByLocation(item.LocationEquipped) {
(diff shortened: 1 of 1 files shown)


jstnmthw:
For added context, I believe this bug stems from the "Back to Town Settings": Equipment Broken (checked). I've experienced the loop as well.

davidfvsilva:
Tested this and it seems to work.

CorncobX:
Also confirmed working with latest compile plus this fix.

Tested with Javazon that would either run out of javelins and facetank mobs if I had "Equipment Broken" unchecked, or it would enter the TP loop with it checked. With this fix, "Equipment Broken" checked now properly goes to town and repairs, then continues the run.

PR #634, opened 2025-01-20
Fix Teleport in circle and Line of sight improvement

elobo91:
- Fix line of sight issues .  Can run tal rasha tombs and stony tombs now =)
- Will reposition if no damage to target for 3 seconds
- Bot wont teleport in circle anymore with lower end cpu ( it wouldnt detect it reached destination so it would teleport once more , get further from destination making a loop/circle) 
- Moved Overshoot logic to utils package   func (pf *PathFinder) BeyondPosition    

Adding RefreshGameData in step move.go loop fix this at minimal cost considering teleport takes 500ms to 1000ms.   so  extra refresh every 500 to 1000ms

### DIFF
--- a/internal/action/item_pickup.go
+++ b/internal/action/item_pickup.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
-	"math"
 	"slices"
 	"time"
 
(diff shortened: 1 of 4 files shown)


sorc653:
Great work, love this fix!

PR #637, opened 2025-01-20
Exit through red portal after Mephisto

Sunblood:


braccali1:
internal/server/templates/run_settings_components.gohtml line 80 checkbox name for ExitToA4 is incorrect, it should be gameMephistoExitToA4 not gameMephistoOpenChests

PR #638, opened 2025-01-21
feature: reopen tp when coming back from town if in leader mode

davidfvsilva:
When in leader mode, sometimes the character goes back to town for various reasons, and then returns to the group but closes the TP leaving others stranded in town. This makes it that the leader opens a fresh portal when going back from town.

### DIFF
--- a/internal/action/town.go
+++ b/internal/action/town.go
@@ -4,6 +4,7 @@ import (
 	"github.com/hectorgimenez/d2go/pkg/data/skill"
 	"github.com/hectorgimenez/koolo/internal/action/step"
 	"github.com/hectorgimenez/koolo/internal/context"
+	"github.com/hectorgimenez/koolo/internal/utils"
 )
 
 func PreRun(firstRun bool) error {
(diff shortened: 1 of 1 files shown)


davidfvsilva:
Let me know if you want it to be a config.

internal/action/town.go
```diff
@@ -95,6 +96,12 @@ func InRunReturnTownRoutine() error {
...
 	Repair()
-
+	
+	if (ctx.CharacterCfg.Companion.Leader) {
+		UsePortalInTown()
+		utils.Sleep(500)
+		return OpenTPIfLeader()
+	}
+	
 	return UsePortalInTown()
```
artosimonyan, line 106: Isn't the second UsePortalInTown() unnecessary hanging outside of the If statement? It should be one or the other. wrap it in an else statement and test if it works both in the leader scenario and non-leader scenario

davidfvsilva: no, because you skip the whole if statement if you are not in Leader mode. You want to either return after opening a TP or return after using the TP.

PR #646, opened 2025-01-26
Show supervisor name in Discord messages

Sunblood:


davidfvsilva:
I'm adding this to the follower coop PR since it's useful when the bots speak in discord. I just changed a bit the formatting, but the idea is there, thanks!

PR #647, opened 2025-01-27
Do town actions only if we made it to town

elobo91:
Has retry mechanism, if still fails after 3 attempt will continue run rather then error out .   InRunReturnTownRoutine()  will be detected by goroutine shortly after.

This fix issue where bot is doing town actions like identifying items, interacting with npc, stashing ( was doing some step dancing) next to portal  in farming area.

### DIFF
--- a/internal/action/town.go
+++ b/internal/action/town.go
@@ -1,6 +1,8 @@
 package action
 
 import (
+	"fmt"
+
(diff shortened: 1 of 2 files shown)


sorc653:
Been running this for a couple days and it's working great!

PR #658, opened 2025-02-03
Refactor move.go, interact_entrance.go, interact_object.go

elobo91:
- New Spirals (EntranceSpiral , ObjectSpiral) using description (.desc) linked to .txt files in d2go .

-  Simplified and cleaned Action Move.go EnsureAreaAsync() no longer required,  removed redundant movement and clicks that doesnt belong there. Prevent an issue with harem/palace cellar (similar area) where there is 2 entrance leading to next area. We pick the nearest.

- Multiple isCompletedFn checks for different objects wich allow to remove EnsureAreaAsync() . 
For portals PlayerUnit.States.HasState(state.JustPortaled)  this is first thing that confirms we interacted with portal, after we check for OpenMenus.LoadingScreen  then finally we check if we reached area. Removed hardcoded  portal destinations since we have them accessible from memory  object.PortalData.DestArea

**interact_entrance.go :**
Now all entrances are compatible with koolo.  It was failing with multiple tal rasha tombs, halls of vaughts and possibly more that i am unaware of. Enhanced using implementation in d2go for Entrances rather then Adjacentlvl coordinates only. Since Adjacentlvl and entrance share exact coordinates (100%) we can link it to unittype5 (entrance) then get the description from lvlwrap.txt  wich contains precious information related to its size,  direction( left, right, bottom) etc. We still need UnitType 2 here because sometimes entrances like ancient tunnel is both (unittype 2 the trap, unittype 5 to enter area)

Simplified runs using red portals : we can directly do action.InteractObject(portal, nil)  and provide no completion check.  On a next update it will be possible to simply do MoveToArea(area.Tristam) for exemple.

cleaned tp_actions.go  and added a check if we stand next to waypoint before creating portal.  Its hard for bot to hover portal due to waypoint being bigger then the portal hitbox. We simply move away from it before making portal.

extra:
Cow lvl:
Configurable option to kill rakanishu( or skip) this speeds up run especially for mosaic who needs to build stack before killing anything. It will go directly in tristam if disabled. Also added a little offset from wirt corpse to move after picking up leg to prevent bot struggling with gold piles


davidfvsilva:
I found a bug with this refactor: I've been testing it.
When you moveToArea in an adjacent level, the bot proceeds to the area BUT he might not have that closest entrance as the destination, and can go ALL THE WAY to the other end of the map to another zone entrance.

davidfvsilva:
I also found an issue with entrances that have animations, like the Kurast temples. Most entrances seem to be fine, including the sewers entrances.

elobo91:
> I also found an issue with entrances that have animations, like the Kurast temples. Most entrances seem to be fine, including the sewers entrances.

Ok i will take a look at both these case.  Looks like will have to check for object mode operating ( wait) then interact with the entrance for these special case . Thanks for testing  !

elobo91:
Reconverting to draft since this pr will highly depend on new pathfinder changes

elobo91:
Closed for rework when AquaRush is done with new pathfinder

PR #660, opened 2025-02-04
fix: stop getting rid of keys after cubing recipes

davidfvsilva:
After buying keys while visiting vendors, the bot can start cubing recipes. After completion of recipes, the bot stashes items that match a nip file entry, and tosses away what doesn't. Unfortunately, this includes tossing away the keys it just bought. This won't happen anymore.


artosimonyan:
Obsolete. Already added with #680

PR #666, opened 2025-02-06
Item Pickup Improvements

braccali1:
Problem: Character was moving/casting during item pickup routine which moves the screen. The coordinates of an item move when the screen moves. This was causing attempts of PickupItem() to fail.
Solution: Added a check if player is casting or moving in PickupItem() function in pickup_item.go action step with a small sleep to minimize cpu usage. Also added this error to ItemPickup() function in item_pickup.go action to not add to the 3 attempts, similar to error ErrMonsterAroundItem and ErrItemTooFar but without debug logging because this check occurs frequently.

Problem: Refreshing game data for hover status was not returning fast enough, so extra iterations of PickupItem() were occurring when the item was actually hovered.
Solution: Add small delay after mouse move and refreshgamedata() that is right before checking hover status in PickupItem() function in pickup_item.go action step. 

Problem: Spiraling too much and timeouts too long. This causes unnecessary delay between retries. Testing showed there is a very low percentage of items successfully being picked up after 6-7 spiral attempts with the above changes, but there was still some success at 8-9 attempts.
Solution: Set max spiraling attempts to 10 (9 in pickup_item.go since it starts at 0). Set the pickupTimeout to not multiple by 2 in the timeout condition check.

Testing showed no blacklisting of items with over 1000 items to pickup, > 90% success on 1st pickup attempt, and an average of 2 spirals per successful pickup attempt. Retries occur much quicker since we are spiraling much less.

EDIT:
Added 2 more retries to the main ItemPickup loop. One following the 2nd and 3rd retry character movement and one that follows the no LOS routine(move beyondposition).

EDIT 2:
Removed item too far from skipping adding to the attempts. There was a loop that could occur if an item was > 7 away but there may have been a door or stash or object in the way. The additional attempts will move more than the 1st or 2nd attempt, and the final attempt executes a movebeyondposition that should resolve the issue.


davidfvsilva:
Tested this PR, and it feels much slower and adds alot of delay to runs. Also, there seems to be a large loop back to go get items that weren't picked up when clearing areas. It seems to have more trouble than usual picking things up (I didn't notice any issues with items being skipped before. The only "problem" I had was not picking up fast enough, but it still felt quick. I wouldn't want to slow it down). I'll give it another go some other times. I tried this PR in p8 terror zones.

braccali1:
@davidfvsilva This PR has been a lot faster than the existing pickup routines in all of my testing. Are you sure there are not any other modifications in your local repo?

davidfvsilva:
> @davidfvsilva This PR has been a lot faster than the existing pickup routines in all of my testing. Are you sure there are not any other modifications in your local repo?

nothing for item pickup on my end, I'll test it again tomorrow on different runs. There just seemed to be some delay between items after the first item pickup when multiples were dropping.

davidfvsilva:
Alright, I'm testing this morning and it's looking ok for one item at a time, but there seems to be a blockage when many things drop at the same time. I'll keep testing to see if I can get more precise info.

davidfvsilva:
Ok, so its actually very quick most of the time and quite reliable. The issue when theres a bunch of items falling on the floor at the same time might not be related to these changes. If anyone else can test, that would be awesome.

braccali1:
@davidfvsilva I found something but it does not fully resolve all pickup attempts yet, but it does make hover checking much much faster and lets spiral attempts also occur faster. refreshgamedata takes a very long time to return ishovered on items, and other objects as well. Calling context.GameReader.GameReader.GetData().HoverData.UnitID against the unitid of the item we want to pickup checks hover without waiting for a refresh of all gamedata to return. You will notice that no item pickups occur on spiral attempt 0 with the current version of this PR. I will work on testing/data gathering of different retry attempt logic and add the new hover logic soon.

tylerqr:
Tested this for around 5-6 hours today. in that time, with these changes, it blacklisted 15 items. See logs attached. For comparison, when I run on main branch w/out these changes, the bot has only blacklisted 1-3 potions after running all night long. 

I did notice these changes make the bot pick up faster generally, but it does seem to blacklist significantly more often. At least in my testing. 
[Supervisor-log-Frizz-2025-02-13-12-07-46.txt](https://github.com/user-attachments/files/18791724/Supervisor-log-Frizz-2025-02-13-12-07-46.txt)

braccali1:
@tylerqr I also got a few black lists but not nearly as many as you. I think it comes down to removing the spiral delay. I highly suggest you empty the gold or gamble on your character. The bot constantly finding gold, attempting to pickup gold, and blacklisting it is probably slowing it down considerably and is also filling the debug log with junk. I am re adding the spiral delay now.

davidfvsilva:
Tested with latest changes, its working well.

artosimonyan:
Looking good so far, I've tested it for 3 hours or so and no blacklists so far. Pickup is quite fast as well. Will test overnight to make sure everything's fine.

When testing, please make sure to test the following areas:
- Cows
- Chaos Sanctuary
- Countess
- Travincal

That should give you a solid testing scenario of all areas that might be problematic. @braccali1, while still working on it and not ready for merge, please mark the PR as draft (work in progress). Mark it back as ready for review when it's ready for merge. 

Feel free to ping me in discord or tag me here when it's ready.

braccali1:
@artosimonyan Converted to draft. Added an item blacklist event. Added the blacklist event to item_pickup.go to screenshot with showitems on and list area.

Testing those areas. Got a blacklist in trav without a line of site or wall problem. I will attempt more work around lots of items in small areas like trav.

braccali1:
I graphed spiral attempts to visualize the effects of adding more attempts.

25 spirals:
Untitled

45 spirals:
Untitled-1

Since we get around 95%+ pickup success on the first 3 attempts, changing the amount of spiral attempts on the 4th and 5th retry should give us better odds while not slowing down the majority of pickup attempts. Also changing the base coordinate 1 or 2 seems to help on 3+ pick up attempt without hitting max spiral attempts. Adding these changes for more testing.

davidfvsilva:
Item Um Rune blacklisted ;_;
image

Most blacklists Ive seen are due to it being on the other side of a wall or structure. Maybe it can do a random movement on a few attempts in order to reposition before the grab?

braccali1:
@davidfvsilva 
Yeah doors and objects are still an issue. I will force movement even if item is within 7 distance on 3+ retries.

artosimonyan:
@braccali1, got reports that there are issues with walking characters. Please test it on a non-teleporting char.

braccali1:
Marking "Ready for review".

Item pickup will never be perfect with current mouse hovering for pickup. Hovering can easily be blocked by objects and other items like chests, shrines, gold, etc..  Not matter how much the character moves or how much spiraling is attempted, there will still be pickup blacklists. In my testing, without an overloaded cpu, this PR works very well.

braccali1:
> @braccali1, got reports that there are issues with walking characters. Please test it on a non-teleporting char.

I did testing with a walking char in trav and it appears to work the same as a teleporting character. The check I added should work with casting running or walking and not add to pickup attempts. Not sure of the exact specifics of the report when walking/running, may have been a fluke or accidentally pressing a modifier key.

PR #671, opened 2025-02-08
Refactor internal/character directory

Cady0:
I am new, go easy on me.

BlizzardSorceress inherits from its parent CharacterBuild which inherits from its parent BaseCharacter. Maybe this is not exactly how the maintainers would like to refactor it, but I hope this *thought* is helpful.

I reduced the character folder size by ~900 lines, but this isn't about saving lines--the main point is that if you expect to add more builds in the future, they all should share common methods such as KillMonster or KillDiablo; that's how inheritance works.

I also noticed that some methods that were copy-pasta'd between all 14 of the character builds has slight differences. Some of these differences were just a comment, some seem trivial, and some I removed, so some functionality is probably broken.

All of the classes would need to be tested and some minor changes are necessary before pulling this, but if you want to keep improving the codebase, more refactoring needs to be done.

### DIFF
--- a/internal/character/character_build.go
+++ b/internal/character/character_build.go
@@ -1,0 +1,227 @@
+package character
+
+import (
+	"fmt"
+	"log/slog"
+	"sort"
+	"time"
+
+	"github.com/hectorgimenez/d2go/pkg/data"
+	"github.com/hectorgimenez/d2go/pkg/data/npc"
+	"github.com/hectorgimenez/d2go/pkg/data/skill"
+	"github.com/hectorgimenez/d2go/pkg/data/stat"
+	"github.com/hectorgimenez/d2go/pkg/data/state"
+	"github.com/hectorgimenez/koolo/internal/action/step"
+	"github.com/hectorgimenez/koolo/internal/game"
+)
+
+const ()
+
+type CharacterBuild struct {
+	BaseCharacter
+}
+
+func (s CharacterBuild) CheckKeyBindings() []skill.ID {
+	missingKeybindings := []skill.ID{}
+	return missingKeybindings
+}
+
+func (s CharacterBuild) KillMonsterSequence(
+	monsterSelector func(d game.Data) (data.UnitID, bool),
+	skipOnImmunities []stat.Resist,
+) error {
+	completedAttackLoops := 0
+	previousUnitID := 0
+	previousSelfBlizzard := time.Time{}
+
+	blizzOpts := step.StationaryDistance(blizzMinDistance, blizzMaxDistance)
+	lsOpts := step.Distance(LSMinDistance, LSMaxDistance)
+
+	for {
+		id, found := monsterSelector(*s.Data)
+		if !found {
+			return nil
+		}
+		if previousUnitID != int(id) {
+			completedAttackLoops = 0
+		}
+
+		if !s.preBattleChecks(id, skipOnImmunities) {
+			return nil
+		}
+
+		if completedAttackLoops >= sorceressMaxAttacksLoop {
+			return nil
+		}
+
+		monster, found := s.Data.Monsters.FindByID(id)
+		if !found {
+			s.Logger.Info("Monster not found", slog.String("monster", fmt.Sprintf("%v", monster)))
+			return nil
+		}
+
+		// Cast a Blizzard on very close mobs, in order to clear possible trash close the player, every two attack rotations
+		if time.Since(previousSelfBlizzard) > time.Second*4 && !s.Data.PlayerUnit.States.HasState(state.Cooldown) {
+			for _, m := range s.Data.Monsters.Enemies() {
+				if dist := s.PathFinder.DistanceFromMe(m.Position); dist < 4 {
+					previousSelfBlizzard = time.Now()
+					step.SecondaryAttack(skill.Blizzard, m.UnitID, 1, blizzOpts)
+				}
+			}
+		}
+
+		if s.Data.PlayerUnit.States.HasState(state.Cooldown) {
+			step.PrimaryAttack(id, 2, true, lsOpts)
+		}
+
+		step.SecondaryAttack(skill.Blizzard, id, 1, blizzOpts)
+
+		completedAttackLoops++
+		previousUnitID = int(id)
+	}
+}
+
+func (s CharacterBuild) killMonster(npc npc.ID, t data.MonsterType) error {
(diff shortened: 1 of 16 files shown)


internal/character/character_build.go
```diff
@@ -0,0 +1,227 @@
...
+			return nil
+		}
+
+		monster, found := s.Data.Monsters.FindByID(id)
+		if !found {
+			s.Logger.Info("Monster not found", slog.String("monster", fmt.Sprintf("%v", monster)))
+			return nil
+		}
+
+		// Cast a Blizzard on very close mobs, in order to clear possible trash close the player, every two attack rotations
```
elobo91, line 63: Eh no .. why all character would get blizzard sorceress logic from what supposed to be base logic?

internal/character/lightning_sorceress.go
```diff
@@ -130,39 +129,6 @@ func (s LightningSorceress) shouldCastStaticField(monster data.Monster) bool {
 	return hpPercentage > LightningStaticFieldThreshold
 }
```
elobo91, line 131: Removing nova specific logic??

internal/character/foh.go
elobo91, line 1: Now how will foh use its own killboss logic? Its removed .

internal/character/nova_sorceress.go
elobo91, line 1: How is it going to use killbosswithstatic now

internal/character/character_build.go
Cady0: Right. That should be removed and have the parent class just use step.PrimaryAttack.

internal/character/lightning_sorceress.go
Cady0: I don't understand the difference other than lightning sorceress's usually have higher -LR than a blizzard sorceress. However, that is irrelevant, because it is very possible that a blizzard sorcress has the same -LR (e.g. wearing Griffon's Eye, lightning Rainbow Facets, or Infinity Runeword). Why does lightning_sorceress.go use magic numbers of 1 and 3 for the distance of static field rather than actual distance of the skill?

Furthermore, why is there a LightningStaticFieldThreshold? Shouldn't static field be cast until it would do less damage than the primary skill?

internal/character/foh.go
Cady0: If FOH truly has special logic for bosses than those can be replaced. I did not test all 14 current builds. Thank you for your review, sir.

internal/character/nova_sorceress.go
Cady0: Similar to Lightning/Fire/Blizzard sorceress, why does Nova Sorceress have a method called KillBossWithStatic but Lightning sorceress doesn't?

Cady0: I notice now that nova and lightning sorcress builds share this method. I just am not quite sure how to implement it in the parent (since I never programmed in Go before), or if that is even what the maintainers would like to do. So I was first just proposing this rough change before committing to a proper cleanup/refactor.

elobo91: Ok when submitting untested things/suggestion  pr can be made as a “draft” to avoid confusion thanks

davidfvsilva:
The main issue is that most character builds tackle situations differently. Every method needs to be overriden. Also, currently, there are a few character builds that are quite generic and don't work very well. You would end up with massive catch all methods in a parent class. What you're probably looking for is the concept of interfaces.

internal/character/character_build.go
```diff
@@ -0,0 +1,227 @@
...
+type CharacterBuild struct {
+	BaseCharacter
+}
+
+func (s CharacterBuild) CheckKeyBindings() []skill.ID {
+	missingKeybindings := []skill.ID{}
+	return missingKeybindings
+}
+
+func (s CharacterBuild) KillMonsterSequence(
```
jasorello, line 29: I also attempted to reduce redundant code, I landed on each build potentially needing its own implementation of KillMonsterSequence and then struggled to make Go pretend to allow for inheritance. I was focusing specifically on sorceress builds, but had updated the appropriate bosses to use static field before heading into KillmonsterSequence.

My vision was to have each individual build just implement this function themselves (for example, a nova sorc wouldn't worry about self blizzard) but I didn't figure out a way to do it that wasn't a massive rewrite.

Cady0:
> The main issue is that most character builds tackle situations differently. Every method needs to be overriden. Also, currently, there are a few character builds that are quite generic and don't work very well. You would end up with massive catch all methods in a parent class. What you're probably looking for is the concept of interfaces.

Disagree respectfully. Most skills function the same: you fire towards an enemy until the enemy dies. If you are outside the range of the skill (e.g. nova or berserk) you move within range.

Kolbot users assign a primary skill--they do not choose a "build"; and Kolbot is tried and proven over years.

The entire strategy of writing special logic for every single character build is not the right approach. The skill FOH does not need special logic, nor does Blizzard. There is no difference between FOH or Blizzard--you should check for immunities to the primary skill then cast.

I noticed there IS a parent class already implemented, but it is in a different folder than the character folder, and it is quite bare. Alas, I gave up. Some one more knowledgeable with the code base should move that parent class to the same folder at least.

davidfvsilva:
> > The main issue is that most character builds tackle situations differently. Every method needs to be overriden. Also, currently, there are a few character builds that are quite generic and don't work very well. You would end up with massive catch all methods in a parent class. What you're probably looking for is the concept of interfaces.
> 
> Disagree respectfully. Most skills function the same: you fire towards an enemy until the enemy dies. If you are outside the range of the skill (e.g. nova or berserk) you move within range.
> 
> Kolbot users assign a primary skill--they do not choose a "build"; and Kolbot is tried and proven over years.
> 
> The entire strategy of writing special logic for every single character build is not the right approach. The skill FOH does not need special logic, nor does Blizzard. There is no difference between FOH or Blizzard--you should check for immunities to the primary skill then cast.
> 
> I noticed there IS a parent class already implemented, but it is in a different folder than the character folder, and it is quite bare. Alas, I gave up. Some one more knowledgeable with the code base should move that parent class to the same folder at least.

Using that same example, paladins need to ensure an aura and a foh can always use conviction for immunities. Also, a FoH can weave holy bolts from the left click, unlike blizzard which requires you to fire it from the right click (which is where a paladin holds his aura). Also, all the other setups don't care about these logic differences, and if it was centralized, you would need to test all builds (and possibly adjust them) each time you would make a configuration change.

Issue #640, opened 2025-01-22
Leveling characters not assigning stat or skillpoints


Wamlad:
Leveling scripts are not assigning stat or skillpoints. This is a known point as it's listed as a to do in the code. i.e. It's not a bug.
However I just wanted to flag the issue on here so any progress or pull requests can be noted.

I'm largely done with implementing the stat allocation myself. So wanted to flag that incase anyone else was going to work on an update.
Newish to coding and new to Go though, so although I've got the stats allocation working, I'd ideally want to implement much better for loops and tidy up / shrink the code.

Wamlad:
Update: 
- Skillspoints is done and working.
- Statpoints I need to build this out better in order to assign all statpoints in one function, but not far off.
I may just submit a separate Pull Request for the skills first. ETA this weekend. Need to tidy it up and ensure it works on all chars before submitting.

Wamlad:
Note: PR 674 now created to solve this.
https://github.com/hectorgimenez/koolo/pull/674

PR #674, opened 2025-02-09
Update leveling_tools.go to add new allocation of stats and skills

Wamlad:
Stat and Skill point allocation updated for leveling characters, when the leveling script option is also set.
**Stats** are done per the d2go order of STR, ENG, DEX and VIT. Which should work well for leveling, reaching gear requirements and mana needs first.
**Skills** are allocated by comparing totals at the current char level. So it's not just allocating skill X at level Y. It takes into account what you've already skilled up.

**Testing**:
- Tested across characters  (sorc & paladin leveling scripts)
- Tested on standard vs. legacy graphics.
- Tested at different char levels, with different stat settings including changing the order, including 0's, ommissions of certain stats.

**Things to note:**
- Several imports (math, slices, sort, strings) were needed to take character config inputs and manipulate them. I have no idea how inefficient this is, but I used them to turn the character settings into usable data in the for loops, no matter the char level, char settings order, ommissions etc.

I'm not a seasoned coder, so would appreciate feedback in making future PRs easy for you devs.
Also appreciate the leveling process in general isn't the main priority but I will be working away at pieces that I use.
This PR solves issue #640

### DIFF
--- a/internal/action/leveling_tools.go
+++ b/internal/action/leveling_tools.go
@@ -57,7 +57,7 @@ var uiSkillRowPositionLegacy = [6]int{110, 195, 275, 355, 440, 520}
 var uiSkillColumnPositionLegacy = [3]int{690, 770, 855}
 
 func EnsureStatPoints() error {
-	// This function will allocate stat points to the character based on the settings in the character configuration file, and in that order.
+	// This function will allocate stat points to the character based on the settings in the character configuration file.
 
 	ctx := context.Get()
 	ctx.SetLastAction("EnsureStatPoints")
@@ -70,7 +70,7 @@ func EnsureStatPoints() error {
 
 		settingsStatPoints := ch.StatPoints() // complete list of statpoints to be leveled per character settings
 
-		// This section sorts the settingsStatPoints map in the order set by the character config files. Go iterates over maps in a random order, so this is necessary.
+		// This section sorts the settingsStatPoints map in the order set by the character config files. Go iterates over maps in a random order, so this is necessary to avoid random ordering.
 		// Collect stats from the map
 		sortedStats := make([]stat.ID, 0, len(settingsStatPoints))
 		for st := range settingsStatPoints {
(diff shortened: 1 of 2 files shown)


Antizerg:
The stat and skill assignment is working nicely, but the current code has us putting only into vita until level 20.  We need to put some into strength asap so we can equip gear.   I recommend at least  34 strength so we can wear a large shield for eventual Ancients Pledge.

Wamlad:
Thanks for testing, and agreed re: strength. I’ve intentionally not amended any of the character scripts though whilst making this compatible. They can be class specific and people can amend their own. Theirs separate work being done on auto level/auto equip also, so that’s separate from this PR.

Antizerg:
> Thanks for testing, and agreed re: strength. I’ve intentionally not amended any of the character scripts though whilst making this compatible. They can be class specific and people can amend their own. Theirs separate work being done on auto level/auto equip also, so that’s separate from this PR.

Good point, I forgot the stat allocation is happening in the specific character go files.

Wamlad:
Note: known issue with the skill allocation order being messed up by skill quests.
i.e. teleport should be allocated at 18, but if we do the den quest, it should be 19.
Will try to amend this to allocate the skill with the highest ID first, so it should allocate more important or new skills first. Instead of leveling nova/fireball etc on leveling chars.

Antizerg:
For some reason Nova isn't being binded under the sorceress_leveling_lightning.go.
Assuming you already fixed this since you seem to be using the sorc leveling script for lightning.

Wamlad:
Yeah the skill binding doesn't seem to work for certain skills, or skills beyond the first 2-3. It also needs updating for the classic graphic settings I think. I'm not sure how the row and column system works so I didn't touch it in this PR though. This is purely the Stat and Skill functions, EnsureStatPoints and EnsureSkillPoints. Tried to keep it minimal for my first PR and try get it into main.

Wamlad:
I split the function into two for loops so it first levels all specified skills once, to ensure pre-requisites are met.
Then allocates skills based on highest ID first, to ensure better skills or teleport are skilled first.
This now fully works whether it's a level 1 char running leveling, a full skill reset allocating all skills, or a char that's part-way leveled already.
Tested on normal and legacy graphics.
Please note, this has nothing to do with the Skill Binding function.
Ready for review. Happy to remove comments that I've left for clearer review, but may be cleaner to remove.

PR #679, opened 2025-02-13
Feature: Rush Run

davidfvsilva:
This run will Rush a game, running through all selected quest. All quests and acts are configurable in the UI. I am currently using this with a rudimentary implementation of Join Online Games which I will share later. There are a few things I want to improve on this but mostly replacing sleeps with events or main player actions, so let me know if you have any suggestions.


artosimonyan:
We previously had an implementation of rushing, just not "translated" into the new way actions work. One thing I expect implemented before considering it for merging is to wait for a rushee (roster member that is not our own character) to enter the TP/zone before continuing. 

Also, this requires logic to be implemented for leader/follower handling in the out of game loop. 

All and all, good job, needs a bit of work but I'm sure we'll get there.

PR #683, opened 2025-02-19
Fix issue with ping ponging between the two entrances to the same area

sorc653:
Sometimes we'll ping pong between the sewers level 1 entrance. I was playing with a new run for palace/harem and noticed I couldn't go from palace level 1 to haram 2 or to palace level 3.

This is because this loop could select the opposing entrance:
`for _, l := range ctx.Data.AdjacentLevels`

I also attempted to sort this by distance but that might be a little too heavy handed.

Just remembering the entrance we're moving the mouse above makes more sense.


Antizerg:
This helps people progress through the act 2 leveling script as well.  Thanks for putting this together

PR #684, opened 2025-02-20
Enable/Repair game.ShiftKey ( shift + left click )

elobo91:
This pr enable Shift Key usage.
ctx.HID.ClickWithModifier(game.LeftButton, screenPos.X, screenPos.Y, game.ShiftKey)

Used for Equipping items or moving back potions from inventory to Belt in correct order if we died.

Original code would always return 0x8000 after doing the comparison, which explains why the modifier behavior wasn't working correctly . **It was reporting the key as pressed regardless of which key was being queried.**


davidfvsilva:
OoooOo I'll put that in the baalrun xD

PR #685, opened 2025-02-20
Update sorceress_leveling.go

Antizerg:
Updated fire leveling Sorc to use firebolt instead of icebolt, allowing sorc to use firebolt starting at level 1.  Also updated the stat selection to get 34 strength first, then pump rest into vita.  This will probably get updated later once auto-equip is implemented


Wamlad:
To not create a bunch of extra PRs, could you please amend the following in this PR to remove TP from being a required keybind?
This stops the error popup on starting a fresh character. The logic is already in other scripts that auto buys the tome and auto binds it.

update:
`requireKeybindings := []skill.ID{skill.TomeOfTownPortal}`
to:
`requireKeybindings := []skill.ID{}`

Antizerg:
> To not create a bunch of extra PRs, could you please amend the following in this PR to remove TP from being a required keybind? This stops the error popup on starting a fresh character. The logic is already in other scripts that auto buys the tome and auto binds it.
> 
> update: `requireKeybindings := []skill.ID{skill.TomeOfTownPortal}` to: `requireKeybindings := []skill.ID{}`

Completed, thanks for the feedback

PR #686, opened 2025-02-20
Introduce a palace run

sorc653:
This run will go from palace 1 => harem 2 => palace 1 => palace 2 => palace 3.

I added this as I specifically wanted to farm small charms with all res (yes these can drop later, but the item level starts dropping at [harem 2](https://diablo.fandom.com/wiki/Area_Level)).

Blocked by #683. This will not work without it.

### DIFF
--- a/internal/run/palace.go
+++ b/internal/run/palace.go
@@ -1,0 +1,93 @@
+package run
+
+import (
+	"github.com/hectorgimenez/d2go/pkg/data"
+	"github.com/hectorgimenez/d2go/pkg/data/area"
+	"github.com/hectorgimenez/koolo/internal/action"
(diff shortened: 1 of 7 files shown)


internal/run/palace.go
```diff
@@ -0,0 +1,93 @@
...
+		ctx: context.Get(),
+	}
+}
+
+func (a Palace) Name() string {
+	return string(config.PalaceRun)
+}
+
+func (a Palace) Run() error {
+	openChests := a.ctx.CharacterCfg.Game.AncientTunnels.OpenChests
```
davidfvsilva, line 26: Use the configs you added :)

PR #687, opened 2025-02-20
feature: better hammerdin behavior, and fixed boss kill issue

davidfvsilva:
Look at him go!

### DIFF
--- a/internal/character/hammerdin.go
+++ b/internal/character/hammerdin.go
@@ -69,7 +69,7 @@ func (s Hammerdin) KillBossSequence(
 		// Reposition right under for perfect hammer hit
 		if previousUnitID == int(id) {
 			if monster.Stats[stat.Life] > 0 {
-				s.PathFinder.RandomMovement()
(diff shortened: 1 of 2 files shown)


yaboso:
This is a huge improvement, thank you very much. The only thing I sometimes notice is that with larger monster packs, teleport is active instead of concentration, even though the bot doesn't try to teleport. Travi is a good spot to reproduce that behaivor

davidfvsilva:
> This is a huge improvement, thank you very much. The only thing I sometimes notice is that with larger monster packs, teleport is active instead of concentration, even though the bot doesn't try to teleport. Travi is a good spot to reproduce that behaivor

I just took a look at it to make sure I didn't miss anything. And it does do the ensure aura step. Now what could be happening is that it will launch that one hammer that kills a weakling, and then teleport to the next so fast that it could be hard to see OR the aura isn't activated long enough for it to apply.

I'll do some more testing this weekend to see if I can improve something. I love running the hammerdin as a follower of my bowazaon, so should be fun to test it out xD

PR #690, opened 2025-02-22
Update terror_zone.go

Timmynat0r:
Missing CatacombsLevel4 in area.Cathedral


Wamlad:
You're missing ", area" by the way.
Haven't tested and don't know if they change is needed but it certainly won't work as written.

Timmynat0r:
> You're missing ", area" by the way. Haven't tested and don't know if they change is needed but it certainly won't work as written.

thx :)

PR #695, opened 2025-02-22
bugfix: Fix crash on first time character setup AND feature: Better configuration for blizzard sorc

jasorello:
Supposedly this PR:
- Ports static field implementation from nova sorc to blizzard sorc
- exposes blizzard sorc config items to character config
- fixes schedule-related bug on first time config-setup by adding defaults to template/config.yaml

I originally started this change to understand how the bot worked, and I _thought_ that the blizzard sorc wasn't using static field at all. My goal was to make the blizzard sorceress use static field, and also expose some of the variables that I had to modify myself while leveling so that they are part of the config instead of hardcoded.

I now know that there was some support for static field, but I think what's in this PR is easier to understand and use without building.


internal/server/assets/js/character_settings.js
```diff
@@ -152,15 +163,21 @@ document.addEventListener('DOMContentLoaded', function () {
...
         const selectedDifficulty = document.getElementById('gameDifficulty').value;
-        updateBossStaticThresholdMin(selectedDifficulty);
-        handleBossStaticThresholdChange();
+        updateBossStaticThresholdMin(selectedDifficulty, 'novaBossStaticThreshold');
+        handleBossStaticThresholdChange('novaBossStaticThreshold');
     }
     
-    function updateBossStaticThresholdMin(difficulty) {
-        const input = document.getElementById('novaBossStaticThreshold');
+    function updateBossStaticThresholdMin(difficulty, elementId) {
```
jasorello, line 179: refactored this and handleBossStaticThresholdChange so i could reuse the method for both static field configurations

davidfvsilva:
https://github.com/hectorgimenez/koolo/pull/691

jasorello:
I didn't know exactly what you meant by linking your PR, but I pushed some updates to accommodate the spirit of that PR

config/template/config.yaml
```diff
@@ -33,11 +51,20 @@ inventory:
   beltColumns: [healing, healing, mana, rejuvenation] # 4 values, each represents the belt column type, allowed values: healing, mana, rejuvenation
 
 character:
-  class: sorceress # Allowed values: sorceress, lightning, hammerdin, foh, paladin (leveling only)
+  class: blizzardsorceress # Allowed values include: blizzardsorceress, lightning, hammerdin, foh, paladin (leveling only)
   useMerc: true
   stashToShared: false
   useTeleport: true # If set to false, bot will not use teleport skill and will walk to the destination
-
+  sorceress: # core sorceress configurations
```
jasorello, line 58: these are configuration items that seem to be shared across the various sorceress builds. We could let them leverage the same configuration, assuming people don't switch the build for a given character often, requiring them to tweak these each time.

internal/character/blizzard_sorceress.go
```diff
@@ -86,6 +119,9 @@ func (s BlizzardSorceress) KillMonsterSequence(
 			s.Logger.Info("Monster not found", slog.String("monster", fmt.Sprintf("%v", monster)))
 			return nil
 		}
+		if monster.Stats[stat.Life] <= 0 {
```
jasorello, line 122: Adding this allows us to stop iterating early when a monster has a length death sequence. Are there downsides to this?

internal/character/blizzard_sorceress.go
```diff
@@ -97,28 +133,19 @@ func (s BlizzardSorceress) KillMonsterSequence(
 			}
 		}
 
-		if s.Data.PlayerUnit.States.HasState(state.Cooldown) {
+		for s.Data.PlayerUnit.States.HasState(state.Cooldown) {
```
jasorello, line 136: Keep checking to see if we're on cooldown and keep using the primary until then, this is to avoid preemptive attackloops where we move on to blizzard but can't actually cast it yet.

I added the sleep to make sure we don't just slam in leftClicks faster than the game can perform them. is 50ms a good number?

davidfvsilva:
> I didn't know exactly what you meant by linking your PR, but I pushed some updates to accommodate the spirit of that PR

mostly to show that I also thought configs should be different for different types of sorcs.

internal/bot/supervisor.go
```diff
@@ -110,6 +110,11 @@ func (s *baseSupervisor) ensureProcessIsRunningAndPrepare() error {
 
 func (s *baseSupervisor) logGameStart(runs []run.Run) {
 	runNames := ""
+	if len(runs) == 0 {
+		message := "No runs were selected, please check your configuration"
+		s.bot.ctx.Logger.Error(message)
+		panic(message)
```
jasorello, line 116: Is it appropriate to bail here? I was trying to find the best way to catastrophically die but give a good error message to the logs. Should we kill d2 as well?

jasorello:
@davidfvsilva  Idk if you like the approach here of the shared common config for sorceresses, but I could update the other builds to also leverage it, if you think it's a good direction. (I don't know what the procedure for merging or choosing what to merge is, here)

Wamlad:
I agree some central config is useful, but should probably be kept to a minimum to not clog the settings page.
This is just me but I'd propose moving a lot back into the underlying scripts:
**Central Config:**
- Use static
Similar to using teleport setting, it's a pretty fundamental change on how the bot operates and how close it gets to mobs/bosses.

**Underlying Character Config/Scripts:**
- attack loops (not sure why this is important enough for main settings?)
- min/max distances (this should be set for each build in the script, is it important enough to have as a variable in the main settings? The only char with massive static distance would be a nova sorc, but they need to be close range for nova anyway)
- boss static threshold (this should simply change based on difficulty, no? If someone wants to min/max their -light res char they can themselves.)

jasorello:
I agree on not wanting to clog up settings too much, I did figure out how to make it collapsible but it's still overwhelming.

Regarding your points:
- `use static`, sure, I can retain this one
- attack loops I adjusted because I was hitting too many loops and had to update the code and build it myself. Looking at Help there were a number of others with the same issue, although not limited to blizzard sorc. I think this is actually related to how the main attack loop iterates and computer speed. Maybe my computer was just looping faster than expected and hitting 40 early? It might be nice to expose this at the high level, but also if the attack loops are fixed maybe it's a non issue.
- min/max: that's a good point, they're generally specific to the builds too. What do you think about exposing the configuration options via the config.yaml with defaults, but not in the UI? That way, someone barely savvy can update the values without doing a rebuild, but it doesn't make the UI messy.
- re: boss threshold, that's a great idea.

davidfvsilva:
> builds

eventually hector takes a look at it and things get merged once in a while. I'll take a look at it again tomorrow, seems like you've been updating this PR alot.

jasorello:
sorry! I keep finding alternative ways to do things that seem better. I think it's in a pretty good spot right now and I'll leave it alone unless there's specific feedback

- More intuitive configuration for blizzard sorc, especially for those who aren't going to grab the code and build it themselves
- Fix the first-time save error for configurations related to the save-schedule code
- An attempt to make the attackCount pattern in blizzard sorc increment more correctly (although I think it's still over-incrementing

config/template/config.yaml
```diff
@@ -33,10 +51,28 @@ inventory:
   beltColumns: [healing, healing, mana, rejuvenation] # 4 values, each represents the belt column type, allowed values: healing, mana, rejuvenation
 
 character:
-  class: sorceress # Allowed values: sorceress, lightning, hammerdin, foh, paladin (leveling only)
+  class: blizzardsorceress # Allowed values include: blizzardsorceress, lightning, hammerdin, foh, paladin (leveling only)
   useMerc: true
   stashToShared: false
   useTeleport: true # If set to false, bot will not use teleport skill and will walk to the destination
+  
+  sorceress: # generic sorceress configurations
```
davidfvsilva, line 59: This file is for general configuration, and shouldn't be used for character specific config or else it can start getting dirty quick. Each character type has his configs.

internal/action/buff.go
```diff
@@ -73,7 +73,7 @@ func Buff() {
 			ctx.HID.PressKeyBinding(kb)
 			utils.Sleep(180)
 			ctx.HID.Click(game.RightButton, 640, 340)
-			utils.Sleep(100)
+			utils.Sleep(200)
```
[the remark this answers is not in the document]
davidfvsilva: if you want to make this a parameter, you could, but just lowering it for **porverty spec** reasons doesn't feel good enough.

internal/character/blizzard_sorceress.go
```diff
@@ -14,20 +15,17 @@ import (
...
-	LSMaxDistance           = 15 // Left skill
-)
-
 type BlizzardSorceress struct {
 	BaseCharacter
 }
 
 func (s BlizzardSorceress) CheckKeyBindings() []skill.ID {
-	requireKeybindings := []skill.ID{skill.Blizzard, skill.Teleport, skill.TomeOfTownPortal, skill.ShiverArmor, skill.StaticField}
+	requireKeybindings := []skill.ID{skill.Blizzard, skill.Teleport, skill.TomeOfTownPortal, skill.ShiverArmor}
```
davidfvsilva, line 23: There is no reason for any kind of sorc to not have one point in static when its being botted, so you don't need to have a config check for this, and just have it mandatory. Typically, any bot is a PvM build, and any pvm sorcs have static.

internal/character/blizzard_sorceress.go
```diff
@@ -53,6 +51,46 @@ func (s BlizzardSorceress) CheckKeyBindings() []skill.ID {
...
+			return nil
+		}
+
+		bossHPPercent := (float64(boss.Stats[stat.Life]) / float64(boss.Stats[stat.MaxLife])) * 100
+
+		// Pull target threshold from config based on difficulty
+		var targetThreshold int
+		switch s.Data.CharacterCfg.Game.Difficulty {
+		case difficulty.Normal:
+			targetThreshold = s.Data.CharacterCfg.Character.Sorceress.StaticFieldThreshold.Normal
```
davidfvsilva, line 67: These values are constants in the game, they dont need to be configured. You can have fixed values and save the runner some headaches. I added a boss kill sequence in the Frozen Orb build (it can probably be improved), bosses usually dont go much down after 60% and then you gotta get that life to 0. Maybe we can work something there, but keep in mind that kill monster sequence bails out, and when you target a boss you don't want to bail out.

internal/character/blizzard_sorceress.go
```diff
@@ -155,6 +194,10 @@ func (s BlizzardSorceress) KillCountess() error {
 }
 
 func (s BlizzardSorceress) KillAndariel() error {
+	if s.Data.CharacterCfg.Character.Sorceress.UseStaticField {
+		return s.killMonsterWithStatic(npc.Andariel, data.MonsterTypeUnique)
```
davidfvsilva, line 198: Kill MonsterByName should just call your get monster by static, because those will be bosses, and no sorc won't have static.

davidfvsilva (changes requested):

config/template/config.yaml
jasorello: Do you think any of these settings should remain here? Or plop them back as constants in `blizzard_sorceress.go`?

I saw the message below about always using static field, I'll strip that bit out, as well as just set constant targets for static field.

max_attacks_loop might be nice to keep in here because it seems like a common complaint in the Help section, but I think that's CPU/speed based. Maybe there's a way to make the "attack loop" only iterate when an attack has actually been performed? Then a sane number could be used here.

internal/action/buff.go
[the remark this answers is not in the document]
jasorello: that's fair. I'm not sure how I'd expose this as a parameter at the moment without it feeling really weird, I'll just revert.

internal/character/blizzard_sorceress.go
jasorello: noted, i'll re-simplify

internal/character/blizzard_sorceress.go
jasorello: I had it configurable because the nova-sorceress had it configurable and I thought that's what some people might want. I agree that 99% of people don't want to think about it and want it to just work, I'll switch these to fixes values

internal/character/blizzard_sorceress.go
jasorello: Ah. I don't want to fire off too many wasted attack loops but I also don't want spam commands. I'll see if there's a safer way to wait for the primary attack to complete before attempting another. That's good background, thanks!

internal/character/blizzard_sorceress.go
jasorello: I noticed in the nova sorceress that some of the bosses didn't get the static treatment: (countess, summoner, pindle, nith) and I kept that distinction here. Do you think all bosses should be static'd?

Or are you just reiterating that we should just always assume that the socreress has static field and should use it when appropriate?

internal/action/buff.go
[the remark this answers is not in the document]
jasorello: This is the wrong spot to fix this problem, it's after the delay. I thought the issue was that we attempted to move before finishing the frozen armor cast but actually it's that we're running through this code before we're able to cast anything.

On my end I can recreate the problem by starting in A4 and going to kill meph as my first move. The "Post CTA Buffing..." message shows while I'm still on the loading screen.

However, if I start in A3 it waits for the loading screen 
```
time=22:33:33 level=DEBUG msg="Loading screen detected. Waiting for game to load before buffing..."
```

I'll revert this back to 100 and take a look at the check for loading screen to see if I can reproduce an issue. I'll leave that saga for another branch, though.

internal/bot/supervisor.go
jasorello: Ok, i'll pull this out for now and if I find a better spot I'll contribute in a different PR

jasorello:
I've somehow screwed up the diff and files are showing up with all lines changed, maybe because of line endings? not sure but I'm going to close this PR and reopen it with more targeted changes and a fixed diff.

PR #698, opened 2025-02-25
Leveling fixes

CarlPoppa1:
- Moved runs to either already existing boss runs or already existing functions within the quest run
- Fixed other issues within leveling_act1.go to leveling_act5.go
- Moved option to use thawing/antidote potions to Duriel and Andariel main boss runs
- Added option to clear levels of Forgotten Tower in Countess run (we can't get to level 5 of the tower by running past mobs before we can teleport)
- Changed some run logic in act 1 leveling script (Den of Evil til 3, Cold Plains til 6, Stony Field til 9, Countess til 14 then Andariel)
- Fixed issue with traversing areas and discovering waypoints erroring if we have to return to town (needed for act 5)

I've deliberately left out the changes I made to open doors more reliably. So this PR will only work smoothly once #697 has been merged. Both PRs will need to be tested together.

More changes to come...

### DIFF
--- a/internal/action/item_pickup.go
+++ b/internal/action/item_pickup.go
@@ -181,7 +181,7 @@ func GetItemsToPickup(maxDistance int) []data.Item {
 
 	for _, itm := range ctx.Data.Inventory.ByLocation(item.LocationGround) {
 		// Skip itempickup on party leveling Maggot Lair, is too narrow and causes characters to get stuck
-		if isLevelingChar && !itm.IsFromQuest() && (ctx.Data.PlayerUnit.Area == area.MaggotLairLevel1 ||
+		if isLevelingChar && itm.Name != "StaffOfKings" && (ctx.Data.PlayerUnit.Area == area.MaggotLairLevel1 ||
 			ctx.Data.PlayerUnit.Area == area.MaggotLairLevel2 ||
 			ctx.Data.PlayerUnit.Area == area.MaggotLairLevel3 ||
 			ctx.Data.PlayerUnit.Area == area.ArcaneSanctuary) {
--- a/internal/run/leveling_act1.go
+++ b/internal/run/leveling_act1.go
@@ -42,21 +42,9 @@ func (a Leveling) act1() error {
 		a.stonyField()
 	}
 
-	if !a.isCainInTown() && !a.ctx.Data.Quests[quest.Act1TheSearchForCain].Completed() {
-		a.deckardCain()
-	}
-
-	// do Tristram Runs until level 14
-	if lvl, _ := a.ctx.Data.PlayerUnit.FindStat(stat.Level, 0); lvl.Value < 14 {
-		a.tristram()
-	}
-
-	// do Countess Runs until level 17
-	if lvl, _ := a.ctx.Data.PlayerUnit.FindStat(stat.Level, 0); lvl.Value < 17 {
-		a.countess()
-	}
-
-	return a.andariel()
+		if !a.ctx.Data.CanTeleport() {
+			a.ctx.CharacterCfg.Game.Countess.ClearFloors = true
+		}
 }
 
 func (a Leveling) bloodMoor() error {
--- a/internal/run/leveling_act1.go
+++ b/internal/run/leveling_act1.go
@@ -3,7 +3,7 @@ package run
 import (
 	"github.com/hectorgimenez/d2go/pkg/data"
 	"github.com/hectorgimenez/d2go/pkg/data/area"
-	"github.com/hectorgimenez/d2go/pkg/data/item"
+	"github.com/hectorgimenez/d2go/pkg/data/difficulty"
 	"github.com/hectorgimenez/d2go/pkg/data/npc"
 	"github.com/hectorgimenez/d2go/pkg/data/object"
 	"github.com/hectorgimenez/d2go/pkg/data/quest"
@@ -23,37 +23,47 @@ func (a Leveling) act1() error {
 
 	running = true
 
-	// clear Blood Moor until level 3
+	// Clear Den of Evil til level 3 - might need to run it in each difficulty if we need more than one respec
 	if lvl, _ := a.ctx.Data.PlayerUnit.FindStat(stat.Level, 0); lvl.Value < 3 {
-		a.bloodMoor()
+		a.ctx.Logger.Debug("Current lvl %s under 3 - Leveling in Den of Evil")
+		return NewQuests().clearDenQuest()
 	}
-
-	// clear Cold Plains until level 6
+	// Do Cold Plains til level 6
 	if lvl, _ := a.ctx.Data.PlayerUnit.FindStat(stat.Level, 0); lvl.Value < 6 {
-		a.coldPlains()
+		return a.coldPlains()
 	}
 
-	if lvl, _ := a.ctx.Data.PlayerUnit.FindStat(stat.Level, 0); lvl.Value == 6 || !a.ctx.Data.Quests[quest.Act1DenOfEvil].Completed() {
-		a.denOfEvil()
-	}
-
-	// clear Stony Field until level 9
+	// Do Stony Field until 9
 	if lvl, _ := a.ctx.Data.PlayerUnit.FindStat(stat.Level, 0); lvl.Value < 9 {
-		a.stonyField()
+		return a.stonyField()
 	}
 
+	// Do Countess Runs until level 14 - skipping in Hell because cold/fire immune - there's no point getting stuck if our merc isn't geared enough yet
+
+	if lvl, _ := a.ctx.Data.PlayerUnit.FindStat(stat.Level, 0); lvl.Value < 14 || a.ctx.Data.CharacterCfg.Game.Difficulty == difficulty.Nightmare {
 		if !a.ctx.Data.CanTeleport() {
 			a.ctx.CharacterCfg.Game.Countess.ClearFloors = true
 		}
-}
-
-func (a Leveling) bloodMoor() error {
-	err := action.MoveToArea(area.BloodMoor)
-	if err != nil {
-		return err
+		return NewCountess().Run()
 	}
 
-	return action.ClearCurrentLevel(false, data.MonsterAnyFilter())
+	if a.ctx.Data.Quests[quest.Act1SistersToTheSlaughter].Completed() {
+		action.ReturnTown()
+		// Do Den of Evil if not complete before moving acts
+		if !a.ctx.Data.Quests[quest.Act1DenOfEvil].Completed() {
+			NewQuests().clearDenQuest()
+		}
+		if !a.isCainInTown() && !a.ctx.Data.Quests[quest.Act1TheSearchForCain].Completed() {
+			NewQuests().rescueCainQuest()
+		}
+
+		action.InteractNPC(npc.Warriv)
+		a.ctx.HID.KeySequence(win.VK_HOME, win.VK_DOWN, win.VK_RETURN)
+
+		return nil
+	} else {
+		return NewAndariel().Run()
+	}
 }
 
 func (a Leveling) coldPlains() error {
--- a/internal/run/leveling_act2.go
+++ b/internal/run/leveling_act2.go
@@ -1,6 +1,8 @@
 package run
 
 import (
+	"fmt"
+
 	"github.com/hectorgimenez/d2go/pkg/data"
 	"github.com/hectorgimenez/d2go/pkg/data/area"
 	"github.com/hectorgimenez/d2go/pkg/data/item"
@@ -23,23 +25,44 @@ func (a Leveling) act2() error {
 	}
 
 	running = true
+
+	if a.ctx.Data.Quests[quest.Act2TheSevenTombs].HasStatus(quest.StatusInProgress5) {
+		action.MoveToCoords(data.Position{
+			X: 5092,
+			Y: 5144,
+		})
+
+		action.InteractNPC(npc.Jerhyn)
+	}
+
+	if a.ctx.Data.Quests[quest.Act2TheSevenTombs].Completed() {
+		action.MoveToCoords(data.Position{
+			X: 5195,
+			Y: 5060,
+		})
+		action.InteractNPC(npc.Meshif)
+		a.ctx.HID.KeySequence(win.VK_HOME, win.VK_DOWN, win.VK_RETURN)
+		return nil
+	}
 	// Find Horadric Cube
 	_, found := a.ctx.Data.Inventory.Find("HoradricCube", item.LocationInventory, item.LocationStash)
 	if found {
 		a.ctx.Logger.Info("Horadric Cube found, skipping quest")
 	} else {
 		a.ctx.Logger.Info("Horadric Cube not found, starting quest")
-		return a.findHoradricCube()
+		return NewQuests().getHoradricCube()
 	}
 
 	if a.ctx.Data.Quests[quest.Act2TheSummoner].Completed() {
 		// Try to get level 21 before moving to Duriel and Act3
 
-		if lvl, _ := a.ctx.Data.PlayerUnit.FindStat(stat.Level, 0); lvl.Value < 18 {
-			return TalRashaTombs{}.Run()
+		if lvl, _ := a.ctx.Data.PlayerUnit.FindStat(stat.Level, 0); lvl.Value < 21 {
+			return NewTalRashaTombs().Run()
 		}
-
-		return a.duriel(a.ctx.Data.Quests[quest.Act2TheHoradricStaff].Completed(), *a.ctx.Data)
+		if a.ctx.Data.Quests[quest.Act2TheHoradricStaff].HasStatus(quest.StatusInProgress3) {
+			a.prepareStaff()
+		}
+		return a.duriel()
 	}
 
 	_, horadricStaffFound := a.ctx.Data.Inventory.Find("HoradricStaff", item.LocationInventory, item.LocationStash, item.LocationEquipped)
@@ -59,55 +82,32 @@ func (a Leveling) act2() error {
 		a.ctx.Logger.Info("Amulet of the Viper found, skipping quest")
 	} else {
 		a.ctx.Logger.Info("Amulet of the Viper not found, starting quest")
-		return a.findAmulet()
+		a.findAmulet()
 	}
 
 	// Summoner
 	a.ctx.Logger.Info("Starting summoner quest")
-	return a.summoner()
-
-}
-
-func (a Leveling) findHoradricCube() error {
-	err := action.WayPoint(area.HallsOfTheDeadLevel2)
+	if a.ctx.Data.Quests[quest.Act2TheSummoner].HasStatus(quest.StatusQuestNotStarted) && !a.ctx.Data.Quests[quest.Act2TaintedSun].HasStatus(quest.StatusQuestNotStarted) {
+		action.InteractNPC(npc.Drognan)
+	}
+	//return error
+	err := NewSummoner().Run()
 	if err != nil {
 		return err
 	}
-
-	err = action.MoveToArea(area.HallsOfTheDeadLevel3)
-	if err != nil {
-		return err
+	// This block can be removed when https://github.com/hectorgimenez/koolo/pull/642 gets merged
+	if _, found := a.ctx.Data.NPCs.FindOne(npc.Summoner); found {
+		return fmt.Errorf("Summoner not dead")
 	}
 
-	err = action.MoveTo(func() (data.Position, bool) {
-		chest, found := a.ctx.Data.Objects.FindOne(object.HoradricCubeChest)
-		if found {
-			a.ctx.Logger.Info("Horadric Cube chest found, moving to that room")
-			return chest.Position, true
-		}
-		return data.Position{}, false
+	a.ctx.Logger.Debug("Moving to red portal")
+	portal, _ := a.ctx.Data.Objects.FindOne(object.PermanentTownPortal)
+
+	action.InteractObject(portal, func() bool {
+		return a.ctx.Data.PlayerUnit.Area == area.CanyonOfTheMagi && a.ctx.Data.AreaData.IsInside(a.ctx.Data.PlayerUnit.Position)
 	})
-	if err != nil {
-		return err
-	}
-
-	action.ClearAreaAroundPlayer(15, data.MonsterAnyFilter())
-
-	obj, found := a.ctx.Data.Objects.FindOne(object.HoradricCubeChest)
-	if !found {
-		return err
-	}
-
-	err = action.InteractObject(obj, func() bool {
-		updatedObj, found := a.ctx.Data.Objects.FindOne(object.HoradricCubeChest)
-		if found {
-			if !updatedObj.Selectable {
-				a.ctx.Logger.Debug("Interacted with Horadric Cube Chest")
-			}
-			return !updatedObj.Selectable
-		}
-		return false
-	})
+	// End of block for removal
+	err = action.DiscoverWaypoint()
 	if err != nil {
 		return err
 	}
@@ -135,19 +135,6 @@ func (a Leveling) findStaff() error {
 	if err != nil {
 		return err
 	}
-	err = action.MoveTo(func() (data.Position, bool) {
-		chest, found := a.ctx.Data.Objects.FindOne(object.StaffOfKingsChest)
-		if found {
-			a.ctx.Logger.Info("Staff Of Kings chest found, moving to that room")
-			return chest.Position, true
-		}
-		return data.Position{}, false
-	})
-	if err != nil {
-		return err
-	}
-
-	action.ClearAreaAroundPlayer(15, data.MonsterAnyFilter())
 
 	obj, found := a.ctx.Data.Objects.FindOne(object.StaffOfKingsChest)
 	if !found {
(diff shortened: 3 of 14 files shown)


Wamlad:
What do you mean by "Stat allocation (with struct to allow for stats to be defined more than once)"? Unsure what you mean by defined more than once. Is that to do with the autoequip?

Some potential fixes on your more changes to come list below. Would like to help out where I can.
Apologies for the wall of text. Can DM to take any discussion out of the PR if preferred.

**Stat allocation** and **Skill allocation** can use PR 674 #674 . It only changes leveling_tools.go so no impact on your changes. It also doesn't require any changes to the character scripts, so should be fine with your autoequip work? Assuming stat points will be amended in those scripts to meet strength requirements etc.

**Autostart without TP bound**, I've tested a solution to this (or a workaround at least). Just need some time to amend it in this PR.. If it's easier for you to add or can improve on the workaround:

- In each of the character leveling scripts, need to amend 'requireKeybindings' to remove 'skill.TomeOfTownPortal'
- In 'leveling_tools.go' remove the line 'skillsToBind = append(skillsToBind, skill.TomeOfTownPortal)' as this was a duplicate, it's already included in the character leveling scripts.
- This will effectively start the game without TPs, but the logic is already built in to buy Tome of TP and ID if it isn't found. So it will just farm the first area until it can. It will then bind the skill too.
- _Bot settings should have return to town for HP and MP unticked._ This isn't ideal, as they'll need to be switched back on later in the Koolo UI. Can that setting be overridden if char is less than level 3 or something? It's triggered in internal/bot/bot.go but I don't know if that setting can be overridden when the bot is running, outside of changing the actual config.yaml and clicking reload configs.
- Only other suggestion is to maybe change the chest opening from false to true for the Den of Evil since it's the first area. Might help get that gold sooner to buy Tomes.

Other suggestion: There is lots of running through Act 1, so when traversing to new areas or even clearing areas I have been repurposing the Andy script section to buy Stamina Potions at the start of these functions (instead of antidote pots). Helps with speed and survivability. Especially running to Countess, 4 pots seemed sufficient if not clearing monsters.

Wamlad:
Cold Plains function needs a quick fix:
image
MoveToArea needs to be directly adjacent, so it can't run this stright from the Rogue Encampment. Whereas WayPoint will traverse all the way to it if the character doesn't have it yet.

CarlPoppa1:
> What do you mean by "Stat allocation (with struct to allow for stats to be defined more than once)"? Unsure what you mean by defined more than once. Is that to do with the autoequip?

Current implementation uses a map with a key of stat.ID to assign stat points. That means we can only apply any single stat in one consecutive action or actions. In reality, we want to assign a few strength points at level 3, a few at level 12 etc to make sure that we can equip better items as we move through the acts. I've moved it to a struct where you input your target points, once that target has been reached it moves on to the next entry in the struct:

`func (s SorceressLeveling) StatPoints() []context.StatAllocation {

	// Define target totals (including base stats)
	targets := []context.StatAllocation{
		{Stat: stat.Vitality, Points: 50},  // Base 10 + 40
		{Stat: stat.Strength, Points: 25},  // Base 10 + 15
		{Stat: stat.Vitality, Points: 65},  // Previous 50 + 15
		{Stat: stat.Strength, Points: 47},  // Previous 25 + 22
		{Stat: stat.Vitality, Points: 999}, // Rest into vit
	}

	return targets
}`

Wamlad:
Sounds like it's all getting there, appreciate the responses. Let me know if I can assist with anything, but happy to test/review when I can.

Re: the stats allocation, is that not what it already does?
Rather than being a list that is progresses through one by one, the only difference (in my mind) is that the current setup shows which level it's triggering the different thresholds. To me this is the exact same thing but more transparent and editable.

The existing one below is from sorceress_leveling_lightning.go. I would change this to get strength earlier, but didn't think it was worth making a PR until autoequip would take advantage of it. But it's very clear for anyone editing what levels these start triggering. So just saying it's something that probably doesn't need an overhaul, and has a working PR if that's something you can save your time on.


`statPoints := make(map[stat.ID]int)

	if lvl.Value < 9 {
		statPoints[stat.Vitality] = 9999
	} else if lvl.Value < 15 {
		statPoints[stat.Energy] = 45
		statPoints[stat.Strength] = 25
		statPoints[stat.Vitality] = 9999
	} else {
		statPoints[stat.Energy] = 60
		statPoints[stat.Strength] = 50
		statPoints[stat.Vitality] = 9999
	}`

Antizerg:
Current script isn't clearing tower floors, which is not ideal for leveling.
I believe you are missing the countess.go update on this PR that you had intended.

The current countess.go on main does not have any action.ClearCurrentLevel.

But we also need to make it a configurable, like you intended so a normal countess run isn't impacted by this.

I believe if you add your local countess.go to this PR, it will fix the issue.

Wamlad:
@Antizerg try running leveling as lightning. Firebolt with no AOE will be far less effective I imagine.
Also was your chickening happening while clearing each floor? Clearing each floor problably helps solve this too.
The actual countess kill will get a lot of chickens early on. So could always have two functions, clear levels 1-4 until level 12, clear all 5 after level 12? The tower levels are narrow and lots of elite/champ packs so its far quicker for leveling in my opinion.

@CarlPoppa1 I know you simplified the runs to match the existing boss runs but I think leveling will sometimes call for a different approach. i.e. Act 1 normal since we have to walk.
I've just finished implementing the ClearThroughPath actionfrom the Diablo script and it's working well for leveling_act1.go.
I suggest adding this instead of the existing Countess boss run, for normal at least.
Using the ClearThroughPath action could also make navigating Barracks, Jail, Catacombs much easier opening doors without being blocked by mobs. If we clear along the way, we won't be far off level 18 & teleport for Act 2.

```
func (a Leveling) countessClearPath() error {
	// Travel to boss level
	err := action.WayPoint(area.BlackMarsh)
	if err != nil {
		return err
	}

	// Move to these areas directly
	action.MoveToArea(area.ForgottenTower)
	action.MoveToArea(area.TowerCellarLevel1)

	// Destination areas to clear towards
	areas := []area.ID{
		area.TowerCellarLevel2,
		area.TowerCellarLevel3,
		area.TowerCellarLevel4,
		area.TowerCellarLevel5, 
	}

	for _, ar := range areas {
		// Get ingame adjacent areas so we can read the exit coordinates, check it matches the next level from script
		adjacentLevels := a.ctx.Data.AdjacentLevels
		nextExitArea := adjacentLevels[0]
		for _, adj := range adjacentLevels {
			if adj.Area == ar {
				nextExitArea = adj
			}
		}
		// Get the exit coordinates (position) and clear towards it
		nextExitPosition := nextExitArea.Position
		action.ClearThroughPath(nextExitPosition, 15, data.MonsterAnyFilter())
		err = action.MoveToArea(ar) // MoveToArea still needed to click exit
		if err != nil {
			return err
		}

	}
	// Countess section from the boss script added, plus ClearThroughPath
	// Try to move around Countess area
	action.MoveTo(func() (data.Position, bool) {
		if areaData, ok := context.Get().GameReader.GetData().Areas[area.TowerCellarLevel5]; ok {
			for _, o := range areaData.Objects {
				if o.Name == object.GoodChest {
					action.ClearThroughPath(o.Position, 30, data.MonsterAnyFilter())
					return o.Position, true
				} // Countess Chest position from 1.13c fetch
			}
		}

		// Try to teleport over Countess in case we are not able to find the chest position, a bit more risky
		if countess, found := a.ctx.Data.Monsters.FindOne(npc.DarkStalker, data.MonsterTypeSuperUnique); found {
			action.ClearThroughPath(countess.Position, 30, data.MonsterAnyFilter())
			return countess.Position, true
		}

		return data.Position{}, false
	})

	// Kill Countess
	return a.ctx.Char.KillCountess()
}
```

Antizerg:
The act 1 go script isn't saving cain properly.  It will go into stoney after the scroll and immediately leave.  It also doesn't right click on it to activate the shrines.
The logs don't show anything as well, but attached is the log.
[Supervisor-log-NewBlizzy-2025-02-28-15-45-42.txt](https://github.com/user-attachments/files/19035451/Supervisor-log-NewBlizzy-2025-02-28-15-45-42.txt)

Wamlad:
Some of the quest scripts no longer work, so den, cain etc. can be skipped or feel free to fix them too :)

Andariel quest will never get completed by the way. This script triggers the main boss farming run instead of the leveling function. It needs that leveling function to go talk to Warriv and get to Act 2.

https://github.com/hectorgimenez/koolo/pull/713
I recommend using this draft PR with walkable logic. Solves door issues and getting stuck on mobs when no teleport. Also helps with exp. ClearThroughPath action not required as this kind of builds it in (when walking only).

https://github.com/hectorgimenez/koolo/issues/718
I lodged this issue, but checking you ran into this too with Rocky Waste?
When first starting act 2 it can't traverse Rocky Waste so constantly quits outside of town. It basically needs DryHills waypoint manually or the bot will never get it.

CarlPoppa1:
> The act 1 go script isn't saving cain properly. It will go into stoney after the scroll and immediately leave. It also doesn't right click on the scroll to activate the shrines. The logs don't show anything as well, but attached is the log. [Supervisor-log-NewBlizzy-2025-02-28-15-45-42.txt](https://github.com/user-attachments/files/19035451/Supervisor-log-NewBlizzy-2025-02-28-15-45-42.txt)

Thanks. Should be fixed by 8a63fb79c7d999fd94ffc69fc530acb5e4e44adf

CarlPoppa1:
> Andariel quest will never get completed by the way. This script triggers the main boss farming run instead of the leveling function. It needs that leveling function to go talk to Warriv and get to Act 2.

I haven't had that problem, have you tried it? It worked for me.

> #718 I lodged this issue, but checking you ran into this too with Rocky Waste? When first starting act 2 it can't traverse Rocky Waste so constantly quits outside of town. It basically needs DryHills waypoint manually or the bot will never get it.

No I haven't had that either

Antizerg:
> #718 I lodged this issue, but checking you ran into this too with Rocky Waste? When first starting act 2 it can't traverse Rocky Waste so constantly quits outside of town. It basically needs DryHills waypoint manually or the bot will never get it.

My fresh level act2 run was able to get to dry hills.  Only thing to note is... I am teleporting.

Antizerg:
> > The act 1 go script isn't saving cain properly. It will go into stoney after the scroll and immediately leave. It also doesn't right click on the scroll to activate the shrines. The logs don't show anything as well, but attached is the log. [Supervisor-log-NewBlizzy-2025-02-28-15-45-42.txt](https://github.com/user-attachments/files/19035451/Supervisor-log-NewBlizzy-2025-02-28-15-45-42.txt)
> 
> Thanks. Should be fixed by [8a63fb7](https://github.com/hectorgimenez/koolo/commit/8a63fb79c7d999fd94ffc69fc530acb5e4e44adf)

I will need to try this on a fresh run, it is probably bugged out now.
It already have scroll in inventory, but this is what it did:
It joined game, went to dark wood to try to get scroll again.  Found tree, then went back to town, talked to Akara, then went to act 2 lol.  I will run a fresh run today and let you know how it goes.

Antizerg:
New Bug: The Jerhyn palace bug after getting act 2 ammy quest.  The current logic doesn't open up palace and it just tries to get in non-stop.  Pausing bot and talking to Drognan solved it.

CarlPoppa1:
> New Bug: The Jerhyn palace bug after getting act 2 ammy quest. The current logic doesn't open up palace and it just tries to get in non-stop. Pausing bot and talking to Drognan solved it.

Its meant to talk to Drognan if the summoner quest isn't started (e.g we haven't reached Arcane sanc) and if the tainted sun quest (Amy) is any status other than not started. That doesn't seem to be working. I'll just get it to talk to Drognan as part of the retrieve ammy function

Antizerg:
New Bug:  It went to summoner, opened tomb and went to canyon, but didn't kill summoner.  It did get canyon WP.  Now it joins new game and just exits.  it rejoins/exits in an endless loop.  Logs don't show any details in troubleshooting this behavior.

Antizerg:
Duriel bug:  It killed Duriel just fine, but when it got to the angel, it didn't talk to him, just stood in front of him

Antizerg:
> @Antizerg try running leveling as lightning. Firebolt with no AOE will be far less effective I imagine. Also was your chickening happening while clearing each floor? Clearing each floor problably helps solve this too. The actual countess kill will get a lot of chickens early on. So could always have two functions, clear levels 1-4 until level 12, clear all 5 after level 12? The tower levels are narrow and lots of elite/champ packs so its far quicker for leveling in my opinion.
> 

I tried lightning, while I do agree charged lightning is great at levels 2-12, it just can't compete with fireball plus the runeword Leaf combo once you hit 12.  Nova is too mana extensive for a poorly geared leveling character, where fireball still wrecks.  

Nova didn't have a chance at killing Andy prior to chickening, Fireball just wrecks her.

Antizerg:
Based on latest files.
Act 2 Bugs:
1. It seems to never acknowledge summoner is completed and will keep running it over and over
2. It doesn't create staff and put it in oriface

Act 3 Bug:
1. It collects all of the organs, makes the will, breaks the altair, but then goes back to spider forest to start collecting organs again.  It never goes into durance to grab WP or go after mephisto

Antizerg:
Act2 bug:
It has happened a few times now, but there is a scenario where both "The Summoner" quest won't be completed, but "The Seven Tombs" is active as well with symbol.  The bot will just join the game and immediately exit it as soon as it determines the quest log.

I think the only way to solve this is to ensure the summoner quest is fully completed before running line 56:
if !a.ctx.Data.Quests[quest.Act2TheSevenTombs].HasStatus(quest.StatusQuestNotStarted) {

CarlPoppa1:
> Act2 bug: It has happened a few times now, but there is a scenario where both "The Summoner" quest won't be completed, but "The Seven Tombs" is active, showing the final tomb symbol. The bot will just join the game and immediately exit it as soon as it determines the quest log.
> 
> I think the only way to solve this is to ensure the summoner quest is fully completed before running line 56: if !a.ctx.Data.Quests[quest.Act2TheSevenTombs].HasStatus(quest.StatusQuestNotStarted) {

You're right. I've fixed it differently though. I've just moved the Duriel check to after the Summoner check. So it'll keep running Summoner until Summoner is killed.

Fujiyama69:
I've not tried this branch yet, but I appreciate your effort.
In my own testing and slow pace of fixing the leveling I did come across a problem where it killed the summoner but TP'd, and exited the run before interacting with the tome and using the waypoint.

Here's my fix for that specific case in case it is relevant and helpful:

```
// We might have killed the summoner already, but we need to finish the quest to get the waypoint
	if a.ctx.Data.Quests[quest.Act2TheSummoner].Completed() &&
		!a.ctx.Data.Quests[quest.Act2ArcaneSanctuary].Completed() {

		a.ctx.Logger.Debug("Summoner already killed, but quest Act 2 Arcane Sanctuary not completed")
		waypointErr := action.WayPoint(area.ArcaneSanctuary)
		if waypointErr != nil {
			a.ctx.Logger.Error("Error waypointing to Arcane Sanctuary", "error", waypointErr)
			return waypointErr
		}
	} else {
		// Start Summoner run to find and kill Summoner
		err = NewSummoner().Run()
		if err != nil {
			a.ctx.Logger.Error("Error running Summoner", "error", err)
			return err
		}
	}

	tome, found := a.ctx.Data.Objects.FindOne(object.YetAnotherTome)
	if !found {
		a.ctx.Logger.Error("Yet Another Tome not found")
		return fmt.Errorf("yet another tome not found")
	}

	// Try to use the portal and discover the waypoint
	err = action.InteractObject(tome, func() bool {
		_, found := a.ctx.Data.Objects.FindOne(object.PermanentTownPortal)
		return found
	})
	if err != nil {
		a.ctx.Logger.Error("Error interacting with Yet Another Tome", "error", err)
		return err
	}

	portal, found := a.ctx.Data.Objects.FindOne(object.PermanentTownPortal)
	if !found {
		a.ctx.Logger.Error("Permanent Town Portal not found")
		return fmt.Errorf("permanent town portal not found")
	}

	err = action.InteractObject(portal, func() bool {
		return a.ctx.Data.PlayerUnit.Area == area.CanyonOfTheMagi && a.ctx.Data.AreaData.IsInside(a.ctx.Data.PlayerUnit.Position)
	})
	if err != nil {
		a.ctx.Logger.Error("Error interacting with Permanent Town Portal", "error", err)
		return err
	}
```

Timmynat0r:
Found Issue when using ensureKeyBinding: true and legacy mode, bot is clicking on belt slot1 and gets stuck with potion in hand

CarlPoppa1:
> Found Issue when using ensureKeyBinding: true and legacy mode, bot is clicking on belt slot1 and gets stuck with potion in hand

Yeah I hadn't pushed my ensureKeyBinding changes. Please can you try again when you get chance? It should be fixed in legacy and normal graphics now. Thanks for testing.

Timmynat0r:
> > Found Issue when using ensureKeyBinding: true and legacy mode, bot is clicking on belt slot1 and gets stuck with potion in hand
> 
> Yeah I hadn't pushed my ensureKeyBinding changes. Please can you try again when you get chance? It should be fixed in legacy and normal graphics now. Thanks for testing.

this fixed the legacy issue thx

I also posted some related issues on #699

Timmynat0r:
Another small suggestion i have, would be the following:

Since you changed the logics for Andariel and Duriel to fit your leveling logic, its hard to merge this pr into other stuff, because you also changed some lines on the original logic. Like in duriel.go:

action.Buff() before Bossfight is now only done if usethawing is active
interacting with duriel entrance is also within the same function, maybe more

They way i fixed those merge issues was to make a copy of duriel.go, named it duriel_leveling.go and change the calls you use now use twice to ....Leveling, in the long run this is maybe not the best idea, because we dont want to blowup the files that are getting used, but for PR testing i find this quite nice because we do not mess up the main logics that way and dont get major merge conflicts.

You can see it in action in my Custom Branch:
https://github.com/Timmynat0r/koolo/tree/Leveling%2BCompanion

Maybe you want to implement it like that for the moment, if not just ignore my wall of text :)

internal/action/item_pickup.go
```diff
@@ -212,7 +212,7 @@ func GetItemsToPickup(maxDistance int) []data.Item {
 
 	for _, itm := range ctx.Data.Inventory.ByLocation(item.LocationGround) {
 		// Skip itempickup on party leveling Maggot Lair, is too narrow and causes characters to get stuck
-		if isLevelingChar && !itm.IsFromQuest() && (ctx.Data.PlayerUnit.Area == area.MaggotLairLevel1 ||
+		if isLevelingChar && itm.Name != "StaffOfKings" && (ctx.Data.PlayerUnit.Area == area.MaggotLairLevel1 ||
```
jdsmi, line 215: probably out of scope here and not your code but why does it matter if this is a leveling character to pick up the staff? 

Shouldn't any run that needs the staff pick it up?

internal/action/leveling_tools.go
```diff
@@ -191,144 +216,168 @@ func EnsureSkillBindings() error {
...
+				utils.Sleep(300)
+			}
+		} else {
+			if _, found := ctx.Data.KeyBindings.KeyBindingForSkill(skill.FireBolt); !found {
+				ctx.Logger.Debug("Lvl 1 sorc found - forcing fire bolt bind")
+				if ctx.GameReader.LegacyGraphics() {
+					ctx.HID.MovePointer(1000, 530)
+				} else {
+					ctx.HID.MovePointer(685, 545)
+				}
```
jdsmi, line 266: I assume this is due to firebolt being a +skill on an item?

If so is there a way to make this reusable so it can be used in the future by other functions? I'm thinking lower res wands or teleport staff/amulet

internal/character/sorceress_leveling_lightning.go
```diff
@@ -206,24 +206,18 @@ func (s SorceressLevelingLightning) SkillsToBind() (skill.ID, []skill.ID) {
...
-		statPoints[stat.Vitality] = 9999
-	} else if lvl.Value < 15 {
-		statPoints[stat.Energy] = 45
-		statPoints[stat.Strength] = 25
-		statPoints[stat.Vitality] = 9999
-	} else {
-		statPoints[stat.Energy] = 60
-		statPoints[stat.Strength] = 50
-		statPoints[stat.Vitality] = 9999
+func (s SorceressLevelingLightning) StatPoints() []context.StatAllocation {
```
jdsmi, line 209: I had a similar thought change by making it just an array of `stat.ID` returned, it looks more verbose than your implementation but has the same function.

Another feature that can be added following your approach is the ability to dynamically set certain stat levels. For example the amount of dexterity required to reach a certain block rate or amount of strength to wear all currently equipped gear (after a respec). Less important now since there's no autogearing but something that would be useful in the future.

internal/character/paladin_leveling.go
```diff
@@ -164,33 +165,18 @@ func (s PaladinLeveling) SkillsToBind() (skill.ID, []skill.ID) {
...
-		statPoints[stat.Vitality] = 220
-		statPoints[stat.Energy] = 0
-	} else {
-		statPoints[stat.Strength] = 86
-		statPoints[stat.Dexterity] = 50
-		statPoints[stat.Vitality] = 300
-		statPoints[stat.Energy] = 0
+	// Define target totals (including base stats)
+	targets := []context.StatAllocation{
+		{Stat: stat.Vitality, Points: 45},  // Base 25 + 20
```
jdsmi, line 172: I think straight points to spend is more clear here than target totals. This of course would mean knowing the base stats of a character class but that's not hard to put together.

jdsmi (commented):
For the stat and skill assignment I made a similar PR (#766) with a similar approach though I didn't modify the original return type of the stat function like you did.

Overall looks SUPER good!

internal/action/leveling_tools.go
```diff
@@ -52,122 +54,145 @@ var uiSkillRowPositionLegacy = [6]int{110, 195, 275, 355, 440, 520}
...
+		for i := 0; i < pointsToSpend; i++ {
+
+			if !spendStatPoint(allocation.Stat) {
+				ctx.Logger.Error(fmt.Sprintf("Failed to spend point in %v", allocation.Stat))
+				continue
+			}
+
+			remainingPoints--
+
+			updatedValue, _ := ctx.Data.PlayerUnit.BaseStats.FindStat(allocation.Stat, 0)
```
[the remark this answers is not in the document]
CarlPoppa1: Yeah, I'll revert these commits and go with your implementation (although there are still conflicting PRs aiming to do the same thing)

internal/action/leveling_tools.go
```diff
@@ -191,144 +216,168 @@ func EnsureSkillBindings() error {
...
 	}
 
-	if len(notBoundSkills) > 0 {
-		ctx.HID.Click(game.LeftButton, ui.SecondarySkillButtonX, ui.SecondarySkillButtonY)
-		utils.Sleep(300)
-		ctx.HID.MovePointer(10, 10)
-		utils.Sleep(300)
+	clvl, _ := ctx.Data.PlayerUnit.FindStat(stat.Level, 0)
+	//Hacky way to find if we're lvling a sorc at clvl 1
+	str, _ := ctx.Data.PlayerUnit.FindStat(stat.Strength, 0)
```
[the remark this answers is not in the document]
CarlPoppa1: Nothing, dunno what I was thinking. I'll change it

internal/run/leveling_act1.go
```diff
@@ -23,79 +19,58 @@ func (a Leveling) act1() error {
 
 	running = true
 
-	// clear Blood Moor until level 3
+	// Clear Den of Evil til level 3 - might need to run it in each difficulty if we need more than one respec
```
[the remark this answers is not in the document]
CarlPoppa1: In my runs I was repeatedly chickening at Treehead Woodfist at the Inifuss Tree. I had much more success going straight to Tower and leveling there.
Countess to 18 is very slow. I think we should always aim to be within 5 levels of each area level, which means we should stop countess at level 12.

I don't understand step 6 in your list but I'm not a fan of a leveling script that needs manual intervention

internal/run/leveling_act1.go
```diff
@@ -23,79 +19,58 @@ func (a Leveling) act1() error {
...
-		a.countess()
-	}
+	if a.ctx.Data.Quests[quest.Act1SistersToTheSlaughter].Completed() {
+		action.ReturnTown()
+		// Do Den of Evil if not complete before moving acts
+		if !a.ctx.Data.Quests[quest.Act1DenOfEvil].Completed() {
+			NewQuests().clearDenQuest()
+		}
+		if !a.isCainInTown() && !a.ctx.Data.Quests[quest.Act1TheSearchForCain].Completed() {
+			NewQuests().rescueCainQuest()
```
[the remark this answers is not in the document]
CarlPoppa1: Thanks. That can be fixed by checking the quest status, I'll have a look

internal/run/leveling_act2.go
```diff
@@ -23,23 +24,32 @@ func (a Leveling) act2() error {
...
 
 	running = true
+
+	if a.ctx.Data.Quests[quest.Act2TheSevenTombs].HasStatus(quest.StatusInProgress5) {
+		action.MoveToCoords(data.Position{
+			X: 5092,
+			Y: 5144,
+		})
+
+		action.InteractNPC(npc.Jerhyn)
```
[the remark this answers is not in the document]
CarlPoppa1: You need to speak to Jerhyn before you speak to Meshif to move onto Act3

internal/action/item_pickup.go
CarlPoppa1: Yeah I guess but I think leveling is the only run that needs the staff, but I guess you're right that no other run will ever go to maggot lair so it doesn't matter

internal/action/leveling_tools.go
CarlPoppa1: Yeah, I haven't bothered properly accounting for +skills or charges on the key binding script but you're right it would be helpful to do that and to move the key binding functions somewhere more generic to be used for non-leveling. It's not a priority for me at the moment though

PR #700, opened 2025-02-25
Chicken settings for scary auras

sorc653:
This adds a few chicken settings so we can nope out if we acquire certain negative auras.

Screenshot 2025-02-25 at 3 50 26 PM

# TODO
* [ ] Conviction
* [ ] Might
* [ ] Fanaticism
* [ ] Decrepify
* [ ] Lower Resist


sorc653:
@hectorgimenez @artosimonyan @elobo91 if I wanted to chicken if mobs are on top of me w/scary auras, would you place that here? I'm nervous about performance in such an important thread.

Instead I could add options to just skip those mobs so you don't consider them to kill?

PR #701, opened 2025-02-26
Wait until fcr/walk time is finished

hectorgimenez:


jasorello:
Is this an approach that could be leveraged in other situations after an action has been performed? (or even bake it in to performing actions?) I was looking at the attack loop in the blizzard sorceress and trying to limit the time that the attackLoop counter incremented without actually performing an attack loop.

One of the things I tried is making the bot retry a primary attack while it waits for blizzard to come off of cooldown. I pulled from the same `Data.PlayerCastDuration()` function to determine what an appropriate wait time is before trying to recast.

hectorgimenez:
Probably we can do the same in other situations and will improve the CPU consumption. Is not the best solution honestly, an sleep will block the application until it finishes, making it less responsive, but for this use case should be acceptable.

PR #702, opened 2025-02-26
Reduce CPU usage (RefreshGameData)

elobo91:
allow one RefreshGameData at a time even if requested by multiple places in code.

ctx.RefreshGameData() : Performs an immediate refresh if it's been more than 20ms since the last one

ctx.RefreshRequest : Signals the background goroutine to reset its timer when it receives a request

The background goroutine continues with the regular 100ms refresh cycle that adapt to RefreshRequest

this is to be used with : https://github.com/hectorgimenez/koolo/pull/701

dont use before the other pr is merged and finished

this also can be paired with  https://github.com/hectorgimenez/d2go/pull/83  for optimal reduction

### DIFF
--- a/internal/bot/bot.go
+++ b/internal/bot/bot.go
@@ -55,14 +55,15 @@ func (b *Bot) Run(ctx context.Context, firstRun bool, runs []run.Run) error {
 		for {
 			select {
 			case <-ctx.Done():
-				cancel()
-				b.Stop()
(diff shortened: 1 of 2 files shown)


elobo91:
closed.  important pr is https://github.com/hectorgimenez/d2go/pull/83  .  this one doesnt seem to help

but 83 reduce by 50%

PR #703, opened 2025-02-26
Added Option for Countess run to clear ghosts

PhoxSillanpaa:
This PR also adds actions/clear_targets.go. The purpose of this addition was to create a function that takes an argument of npc.ID and creates a list of monsters to kill as the bot moves through the rooms of a dungeon. It functions essentially identically to action.ClearCurrentLevel(), also accepting a boolean to toggle opening chests.

There is likely a more elegant solution to this.

### DIFF
--- a/config/template/config.yaml
+++ b/config/template/config.yaml
@@ -170,6 +170,8 @@ game:
       - 119 # Icy Cellar
       - 121 # Nihlathak's Temple (Will do Nihlathak run)
       - 128 # The Worldstone Keep Level 1 (Will do Baal run)
+  countess:
+    clearGhosts: false # Will go room to room in the Tower and kill every ghost.   
 
 companion:
   enabled: false
--- a/internal/action/clear_targets.go
+++ b/internal/action/clear_targets.go
@@ -1,0 +1,118 @@
+package action
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/hectorgimenez/d2go/pkg/data"
+	"github.com/hectorgimenez/d2go/pkg/data/npc"
+	"github.com/hectorgimenez/d2go/pkg/data/stat"
+	"github.com/hectorgimenez/koolo/internal/context"
+	"github.com/hectorgimenez/koolo/internal/game"
+	"github.com/hectorgimenez/koolo/internal/utils"
+)
+
+func ClearCurrentLevelTargets(openChests bool, ID npc.ID) error {
+	ctx := context.Get()
+	ctx.SetLastAction("ClearCurrentLevel")
+
+	rooms := ctx.PathFinder.OptimizeRoomsTraverseOrder()
+	for _, r := range rooms {
+		err := clearRoomID(r, ID)
+		if err != nil {
+			ctx.Logger.Warn("Failed to clear room: %v", err)
+		}
+
+		if !openChests {
+			continue
+		}
+
+		for _, o := range ctx.Data.Objects {
+			if o.IsChest() && o.Selectable && r.IsInside(o.Position) {
+				err = MoveToCoords(o.Position)
+				if err != nil {
+					ctx.Logger.Warn("Failed moving to chest: %v", err)
+					continue
+				}
+				err = InteractObject(o, func() bool {
+					chest, _ := ctx.Data.Objects.FindByID(o.ID)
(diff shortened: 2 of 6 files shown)


elobo91:
> This PR also adds actions/clear_targets.go. The purpose of this addition was to create a function that takes an argument of npc.ID and creates a list of monsters to kill as the bot moves through the rooms of a dungeon. It functions essentially identically to action.ClearCurrentLevel(), also accepting a boolean to toggle opening chests.
> 
> There is likely a more elegant solution to this.

hi thanks for putting time on this . Why do we need a new clear action? Could create a local monster filter with npc ids you need . Can look how its done here : https://github.com/hectorgimenez/koolo/blob/523a6a0a211435284da0f2a6a747dfce8b3c20e4/internal/run/diablo.go#L167

PhoxSillanpaa:
> > This PR also adds actions/clear_targets.go. The purpose of this addition was to create a function that takes an argument of npc.ID and creates a list of monsters to kill as the bot moves through the rooms of a dungeon. It functions essentially identically to action.ClearCurrentLevel(), also accepting a boolean to toggle opening chests.
> > There is likely a more elegant solution to this.
> 
> hi thanks for putting time on this . Why do we need a new clear action? Could create a local monster filter with npc ids you need . Can look how its done here :
> 
> https://github.com/hectorgimenez/koolo/blob/523a6a0a211435284da0f2a6a747dfce8b3c20e4/internal/run/diablo.go#L167

Thanks. _That_ was the more elegant solution I was looking for. I totally understand how that works. Shame I just spent several hours doing basically what I did for this Countess Run to the Arcane Sanctuary run. But I should be able to rework it easy enough.

Wamlad:
Not sure how to add this in with your Ghosts filter sorry, but I've been testing the leveling scripts and added ClearThroughPath to the leveling countess run. I saw Elb mentioned it in Discord to you so hopefully this is helpful. I think the leveling runs will sometimes inherently be different from the main runs, but this could be worth adding here too if you can build it in with your ghosts filter.
It could be made configurable in the settings separately, but probably just part of your clearGhosts option. Only leveling chars would want to clear everything on the way anyway.

```
func (a Leveling) countessClearPath() error {
	// Travel to boss level
	err := action.WayPoint(area.BlackMarsh)
	if err != nil {
		return err
	}

	// Move to these areas directly
	action.MoveToArea(area.ForgottenTower)
	action.MoveToArea(area.TowerCellarLevel1)

	// Destination areas to clear towards
	areas := []area.ID{
		area.TowerCellarLevel2,
		area.TowerCellarLevel3,
		area.TowerCellarLevel4,
		area.TowerCellarLevel5, //// does this need to be removed? Not if it's just the destination
	}

	for _, ar := range areas {
		// Get ingame adjacent areas so we can read the exit coordinates, check it matches the next level from script
		adjacentLevels := a.ctx.Data.AdjacentLevels
		nextExitArea := adjacentLevels[0]
		for _, adj := range adjacentLevels {
			if adj.Area == ar {
				nextExitArea = adj
			}
		}
		// Get the exit coordinates (position) and clear towards it
		nextExitPosition := nextExitArea.Position
		action.ClearThroughPath(nextExitPosition, 15, data.MonsterAnyFilter())
		err = action.MoveToArea(ar) // MoveToArea still needed to click exit
		if err != nil {
			return err
		}

	}
	// Countess script added, plus ClearThroughPath
	// Try to move around Countess area
	action.MoveTo(func() (data.Position, bool) {
		if areaData, ok := context.Get().GameReader.GetData().Areas[area.TowerCellarLevel5]; ok {
			for _, o := range areaData.Objects {
				if o.Name == object.GoodChest {
					action.ClearThroughPath(o.Position, 30, data.MonsterAnyFilter())
					return o.Position, true
				} // Countess Chest position from 1.13c fetch
			}
		}

		// Try to teleport over Countess in case we are not able to find the chest position, a bit more risky
		if countess, found := a.ctx.Data.Monsters.FindOne(npc.DarkStalker, data.MonsterTypeSuperUnique); found {
			action.ClearThroughPath(countess.Position, 30, data.MonsterAnyFilter())
			return countess.Position, true
		}

		return data.Position{}, false
	})

	// Kill Countess
	return a.ctx.Char.KillCountess()
}
```

PR #707, opened 2025-02-27
feature: updates to blizzard sorc attack loop

jasorello:
I had accidentally polluted my commits in https://github.com/hectorgimenez/koolo/pull/695 with bad line endings, this is a cleaned PR with just the blizzard sorc changes

This PR attempts:
1. Better static field usage, porting the more complicated logic over from the nova sorceress character, as well as moving the distance logic to the common config so it can be more easily referenced across all other sorceress builds
2. Moved blizzard-specific configurations to a function within the struct, keeping the same ease of configuration but without polluting the rest of the character package with consts (i also fixed some other cross-character references that were caught up in this)
3. I did some optimization of the blizzard attack loop. The problem I'm trying to fix here is that on a faster machine, we're iterating through this loop much faster than the bot can actually perform any of the attacks. We can hit 200+ attack loops on a boss after just performing 6 or 7 actual attacks within the game. I think the issues here may also be present in other characters, but I'll wait for someone smarter than me to weigh in.

I also included the switch from `sorceress` to `blizzardsorceress` that was suggested in https://github.com/hectorgimenez/koolo/pull/691

### DIFF
--- a/config/template/config.yaml
+++ b/config/template/config.yaml
@@ -33,11 +33,16 @@ inventory:
   beltColumns: [healing, healing, mana, rejuvenation] # 4 values, each represents the belt column type, allowed values: healing, mana, rejuvenation
 
 character:
-  class: sorceress # Allowed values: sorceress, lightning, hammerdin, foh, paladin (leveling only)
+  class: blizzardsorceress # Allowed values include: blizzardsorceress, lightning, hammerdin, foh, paladin (leveling only)
   useMerc: true
   stashToShared: false
   useTeleport: true # If set to false, bot will not use teleport skill and will walk to the destination
 
+  sorceress: # generic sorceress configurations
+    # static field configuration options
+    static_field_minimum_distance: 8
+    static_field_maximum_distance: 12
+    
 game:
   minGoldPickupThreshold: 500000 # If total gold amount is less than this, bot will pick up and sell magic+ items
   clearTPArea: true # Will clear the TP area before clicking it
--- a/internal/character/blizzard_sorceress.go
+++ b/internal/character/blizzard_sorceress.go
@@ -150,10 +150,10 @@ func (s BlizzardSorceress) KillMonsterSequence(
 		}
 
 		for s.Data.PlayerUnit.States.HasState(state.Cooldown) && s.MonsterStillAlive(id) {
-			step.PrimaryAttack(id, 2, true, lsOpts)
+			step.PrimaryAttack(id, 1, true, lsOpts)
 			actionsTakenThisLoop = append(actionsTakenThisLoop, "PrimaryAttack")
 			// Wait for the cast to complete before doing anything else
-			time.Sleep(s.Data.PlayerCastDuration()) // stolen from internal/action/step/attack.go
+			time.Sleep(s.Data.PlayerCastDuration() - (120 * time.Millisecond)) // stolen from internal/action/step/attack.go
 		}
 
 		selfBlizzardThisLoop := false
--- a/internal/character/blizzard_sorceress.go
+++ b/internal/character/blizzard_sorceress.go
@@ -165,7 +165,7 @@ func (s BlizzardSorceress) KillMonsterSequence(
 		}
 
 		completedAttackLoops++
-		s.Logger.Info("Actions taken:", slog.String("actionsTakenThisLoop", fmt.Sprintf("%v", actionsTakenThisLoop)))
+		// s.Logger.Debug("Actions taken:", slog.String("actionsTakenThisLoop", fmt.Sprintf("%v", actionsTakenThisLoop)))
 
 		previousUnitID = int(id)
 	}
--- a/internal/character/blizzard_sorceress.go
+++ b/internal/character/blizzard_sorceress.go
@@ -22,11 +22,9 @@ type BlizzardSorceress struct {
 func (s BlizzardSorceress) attackConfig() map[string]int {
 	// Avoiding constants that pollute the other characters in the package
 	return map[string]int{
-		"maxAttacksLoop":           100,
-		"leftSkillMinimumDistance": 8,
-		"leftSkillMaximumDistance": 20,
-		"blizzardMinimumDistance":  6,
-		"blizzardMaximumDistance":  15,
+		"maxAttacksLoop":    100,
+		"attackMinDistance": 8,
+		"attackMaxDistance": 16,
 	}
 }
 
@@ -107,13 +105,9 @@ func (s BlizzardSorceress) KillMonsterSequence(
 	previousUnitID := 0
 	previousSelfBlizzard := time.Time{}
 
-	blizzOpts := step.StationaryDistance(
-		s.attackConfig()["blizzardMinimumDistance"],
-		s.attackConfig()["blizzardMaximumDistance"],
-	)
-	lsOpts := step.Distance(
-		s.attackConfig()["leftSkillMinimumDistance"],
-		s.attackConfig()["leftSkillMaximumDistance"],
+	attackOpts := step.StationaryDistance(
+		s.attackConfig()["attackMinDistance"],
+		s.attackConfig()["attackMaxDistance"],
 	)
 
 	for {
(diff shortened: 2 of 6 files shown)


config/template/config.yaml
```diff
@@ -33,11 +33,16 @@ inventory:
   beltColumns: [healing, healing, mana, rejuvenation] # 4 values, each represents the belt column type, allowed values: healing, mana, rejuvenation
 
 character:
-  class: sorceress # Allowed values: sorceress, lightning, hammerdin, foh, paladin (leveling only)
+  class: blizzardsorceress # Allowed values include: blizzardsorceress, lightning, hammerdin, foh, paladin (leveling only)
   useMerc: true
   stashToShared: false
   useTeleport: true # If set to false, bot will not use teleport skill and will walk to the destination
 
+  sorceress: # generic sorceress configurations
```
jasorello, line 41: common configuration that can be leveraged via `s.Data.CharacterCfg.Character.Sorceress.StaticFieldMinDist`

internal/character/blizzard_sorceress.go
```diff
@@ -14,20 +15,24 @@ import (
...
-	blizzMaxDistance        = 20
-	LSMinDistance           = 6
-	LSMaxDistance           = 15 // Left skill
-)
-
 type BlizzardSorceress struct {
 	BaseCharacter
 }
 
+func (s BlizzardSorceress) attackConfig() map[string]int {
```
jasorello, line 22: I noticed there were issues in the builds with accidental references across the characters which might have unintended results. I thought this might be a better pattern to make it more clear which settings are relevant for each character. Is there a better way?

internal/character/blizzard_sorceress.go
```diff
@@ -53,6 +58,46 @@ func (s BlizzardSorceress) CheckKeyBindings() []skill.ID {
 	return missingKeybindings
 }
 
+func (s BlizzardSorceress) killMonsterWithStatic(bossID npc.ID, monsterType data.MonsterType) error {
```
jasorello, line 61: This is mostly stolen from the nova sorceress

internal/character/blizzard_sorceress.go
```diff
@@ -53,6 +58,46 @@ func (s BlizzardSorceress) CheckKeyBindings() []skill.ID {
...
+		if !found || boss.Stats[stat.Life] <= 0 {
+			return nil
+		}
+
+		bossHPPercent := (float64(boss.Stats[stat.Life]) / float64(boss.Stats[stat.MaxLife])) * 100
+
+		// Pull target threshold from config based on difficulty
+		var targetThreshold int
+		switch s.Data.CharacterCfg.Game.Difficulty {
+		case difficulty.Normal:
```
jasorello, line 73: These values have a bit of a buffer over the bare minimum for the difficulty to minimize wasted casts

internal/character/blizzard_sorceress.go
```diff
@@ -77,7 +128,8 @@ func (s BlizzardSorceress) KillMonsterSequence(
 			return nil
 		}
 
-		if completedAttackLoops >= sorceressMaxAttacksLoop {
+		if completedAttackLoops >= s.attackConfig()["maxAttacksLoop"] {
+			s.Logger.Error("Exceeded MaxAttacksLoop", slog.String("completedAttackLoops", fmt.Sprintf("%v", completedAttackLoops)))
```
jasorello, line 132: I've noticed a few people in `Help` noting that their character is bailing early but not knowing why. I think many of them are from hitting this loop in their respective character. This log message should help identify those cases better.

internal/character/blizzard_sorceress.go
```diff
@@ -86,39 +138,34 @@ func (s BlizzardSorceress) KillMonsterSequence(
 			s.Logger.Info("Monster not found", slog.String("monster", fmt.Sprintf("%v", monster)))
 			return nil
 		}
+		if monster.Stats[stat.Life] <= 0 {
```
jasorello, line 141: This avoids unnecessary attackloop iterations for bosses with an extended death animation like andariel

internal/character/blizzard_sorceress.go
```diff
@@ -86,39 +138,34 @@ func (s BlizzardSorceress) KillMonsterSequence(
...
+			return nil
+		}
+
+		for s.Data.PlayerUnit.States.HasState(state.Cooldown) {
```
jasorello, line 145: switched this to cast our leftskill until we're out of cooldown instead of just a single time. This should better leverage our activity during blizzard cooldown without resulting in a new attack loop.

I've pulled it ahead of the blizzard self-cast because otherwise we were usually hitting the self-cast block while blizzard was on cooldown.

New behavior should be:
```
while on cooldown:
	cast leftskill
	
if timeToCastSelfBlizzard:
	cast selfBlizzard
else
	cast regularBlizzard:
```

internal/character/blizzard_sorceress.go
```diff
@@ -86,39 +138,34 @@ func (s BlizzardSorceress) KillMonsterSequence(
...
-
-		step.SecondaryAttack(skill.Blizzard, id, 1, blizzOpts)
-
 		completedAttackLoops++
+
 		previousUnitID = int(id)
 	}
 }
 
-func (s BlizzardSorceress) killMonster(npc npc.ID, t data.MonsterType) error {
```
jasorello, line 111: This function was no longer being called

internal/character/blizzard_sorceress.go
elobo91: Hello . Thats a good catch. There isnt any character using this yet but maybe it could be useful in that case? 
IMG_2060

could use npcmode 0 and 12 instead of life stat to determine if its death if its problematic.

elobo91: I meant using mode dead and death . Dead is when it just died, death is when its extremely dead ( after animation)

jasorello:
The blizzard logic is busted in this, that's what I get for doing a refactor late and night and pushing without testing. I've working on a fix so it actually casts offensive blizzards again but I'm still seeing attack loops where nothing is performed.

jasorello:
I ended up pulling out the logic for 'is this monster dead?' into a separate function, i'll see if elobo91's suggestion can make it simpler.

internal/character/blizzard_sorceress.go
jasorello: @elobo91 let's "pretend" i'm a dope: how can I leverage this? I don't see any references in koolo yet and the references in d2go appear to be leveraging memory offsets for the monster's location in memory. Would a function need to be exposed in d2go first?

elobo91: image

elobo91: is exposed

jasorello: i'll investigate, thanks for the details!

jasorello: implemented, thanks!

elobo91: Still have to check wich of the 2 mode is set after the animation otherwize it will still exit game early and have no real benefit from checking life stat

jasorello: ahh gotcha i missed that distinction. i'll think about how I can implement that, i don't think it fits with how I set up the methods at this time, right?

elobo91: Also keep in mind koolo is able to read itemlocation dropping . Both location ground and dropping are merged together. This means it reads item before they are visible on ground. Could explain early exit symptoms if nothing match .nip rule . We dont have to see them on ground to evaluate if it should pickup

jasorello: I've been thinking about what you said and I'm not sure the distinction matters for the attack loop. Either the monster isn't dead and we should kill it, or it's dead and we should do something else.

Can you give some more background on the problem you're describing? Should there be a short wait added in the run definition or in the KillMephisto definition?

jasorello:
merged in main, kept the infinite loop for killmonstersequence and dropped the attackloop max back to where it was before

PR #708, opened 2025-02-27
Bugfix: fail more gracefully when supervisor starts with no runs selected

jasorello:
The current behavior when no runs are selected is to crash the supervisor without a useful error.

This PR does a check before we start the first run to ensure we have at least one run selected, otherwise it kills the supervisor and prints a message that the user hopefully reads before they holler at discord asking why their bot won't start.

Maybe the javascript validation should try to prevent someone getting into this state?


jasorello:
I realized this has a bug that means the runs never get refreshed as expected and there also doesn't seem to be an appetite for the better error. Happy to contribute an alternative fix in the future.

PR #713, opened 2025-02-27
Walkable logic ( door handling and monsters)

elobo91:


jasorello:
I love the clearing ability, I'm trying this for the first time on the diablo run and find that it gets caught up with "monsters in the path" when we're near the start or terminus of a path. For example, if you're doing a regular clear, if there's a monster where it wants to be at the entrance it will often bail. Or, if you're "starting from the star" (but still walking) it will clear to the star and then bail.

Not sure what to tweak to adjust this, but wanted to share the feedback.

jasorello:
Another space I see the bail error consistently is on a walking cows run. When going for wirt's leg it'll get caught up on the champ pack that's always northwest of cain's cage and bail.

elobo91:
> > I love the clearing ability, I'm trying this for the first time on the diablo run and find that it gets caught up with "monsters in the path" when we're near the start or terminus of a path. For example, if you're doing a regular clear, if there's a monster where it wants to be at the entrance it will often bail. Or, if you're "starting from the star" (but still walking) it will clear to the star and then bail.
> > Not sure what to tweak to adjust this, but wanted to share the feedback.
> > ```
> > time=22:24:36 level=WARN msg="Game finished with errors, reason: monsters detected in movement path. Game total time: 146.48s" supervisor="me" mapSeed=1019905429
> > ```
> 
> Looks like this is caused by the use of ClearThroughPath in the Diablo.go, if you change those to MoveToCoords, it will stop exiting. While this solves the issue of the bot bailing, it appears that it will just run circles around the seals until the end of time, so it's probably not a fix for Diablo, but might help you get cows going with this

yes, diablo is currently on script that use  ClearThroughPath.   since it handle movement in segments it creates conflict.  Need to disable  clearPath from  move.go for walkable character when  using ClearThroughPath.  Or workaround  add exception for chaos sanctuary  to simply  return step.moveto   like walkable character.

this pr is a draft .  i didnt have time to rework this a lot should change.  i left it open so someone else could start from it and possibily improve/finish it.

ive been using only little part of this pr on reset  with walkable FOH  pits  then moved with walkable chaos runs until i had enigma haha.

jdsmi:
I've been testing this out and it's much much better now. There are a still breaking quirks though:

1. Often when going through entrances with or without a door if the character is hugging the wall near it they'll just pace back and forth against the wall near the door. This is most detrimental when in areas with lots of twists and turns and of course doors to open like the barracks/jail/catacombs/act 2 temples/etc. Sometimes the character gets through but mostly they just pace unti it gives up and restarts the game
2. Some doors have rare chances to even work, eg the tomb of tal rasha. It'll just fail attempts to use it then restart.
3. the clearing part is great but it seems to blacklist monsters a bit over aggressively. Often it'll attack something, blacklist it for 30 seconds then stand there getting punched in the face until the blacklist times out or it chickens

elb91:
> I've been testing this out and it's much much better now. There are a still breaking quirks though:
> 
> 1. Often when going through entrances with or without a door if the character is hugging the wall near it they'll just pace back and forth against the wall near the door. This is most detrimental when in areas with lots of twists and turns and of course doors to open like the barracks/jail/catacombs/act 2 temples/etc. Sometimes the character gets through but mostly they just pace unti it gives up and restarts the game
> 2. Some doors have rare chances to even work, eg the tomb of tal rasha. It'll just fail attempts to use it then restart.
> 3. the clearing part is great but it seems to blacklist monsters a bit over aggressively. Often it'll attack something, blacklist it for 30 seconds then stand there getting punched in the face until the blacklist times out or it chickens

https://discord.com/channels/1163042152985661480/1353138361715654758

Discussion about this

jdsmi:
> > > I love the clearing ability, I'm trying this for the first time on the diablo run and find that it gets caught up with "monsters in the path" when we're near the start or terminus of a path. For example, if you're doing a regular clear, if there's a monster where it wants to be at the entrance it will often bail. Or, if you're "starting from the star" (but still walking) it will clear to the star and then bail.
> > > Not sure what to tweak to adjust this, but wanted to share the feedback.
> > > ```
> > > time=22:24:36 level=WARN msg="Game finished with errors, reason: monsters detected in movement path. Game total time: 146.48s" supervisor="me" mapSeed=1019905429
> > > ```
> > 
> > 
> > Looks like this is caused by the use of ClearThroughPath in the Diablo.go, if you change those to MoveToCoords, it will stop exiting. While this solves the issue of the bot bailing, it appears that it will just run circles around the seals until the end of time, so it's probably not a fix for Diablo, but might help you get cows going with this
> 
> yes, diablo is currently only script that use ClearThroughPath. since it handle movement in segments it creates conflict. Need to disable clearPath from move.go for walkable character when using ClearThroughPath. Or workaround add exception for chaos sanctuary to simply return step.moveto like walkable character. Other runs wont have this issue since they use cleararea
> 
> this pr is a draft . i didnt have time to rework this a lot should change. i left it open so someone else could start from it and possibily improve/finish it.
> 
> ive been using only little part of this pr on reset with walkable FOH pits then moved with walkable chaos runs until i had enigma haha.

I had this happening to me doing non leveling runs. I found this actually happens when the `action.MoveTo` function is used instead of the `step.MoveTo` that this PR greatly improves.

The `action.MoveTo` is what returns the error so when used directly it needs to be handled accordingly.

`step.MoveTo` however does address that error and, as this PR does wonderfully, clears the monsters

internal/action/step/move.go
```diff
@@ -146,3 +184,66 @@ func MoveTo(dest data.Position, options ...MoveOption) error {
...
+			doorPos := o.Position
+			ourPos := ctx.Data.PlayerUnit.Position
+
+			// Calculate if door is roughly in our path to destination
+			dotProduct := (doorPos.X-ourPos.X)*(dest.X-ourPos.X) + (doorPos.Y-ourPos.Y)*(dest.Y-ourPos.Y)
+
+			lengthSquared := (dest.X-ourPos.X)*(dest.X-ourPos.X) + (dest.Y-ourPos.Y)*(dest.Y-ourPos.Y)
+
+			// If door is roughly in our direction of travel and close enough
+			if lengthSquared > 0 && dotProduct > 0 && dotProduct < lengthSquared {
```
klb92, line 206: I tried calculating with the cross and it seems to have a much better time with the doors in jail etc.

I changed the function to :

`func handleObstaclesInPath(dest data.Position, openedDoors map[object.Name]data.Position) bool {
	ctx := context.Get()

	// Check for doors in the path
	for _, o := range ctx.Data.Objects {
		if o.IsDoor() && o.Selectable &&
			ctx.PathFinder.DistanceFromMe(o.Position) < 5 &&
			openedDoors[o.Name] != o.Position {

			// Check if door is between us and destination
			doorPos := o.Position
			ourPos := ctx.Data.PlayerUnit.Position


			// A threshhold so we open the door even if the door is not directly in our path, but close enough
			threshhold := 5.0

			// If door is roughly in our direction of travel and close enough
			if isObjectInPath(dest, ourPos, doorPos, threshhold) {
				ctx.Logger.Debug("Door detected in path, opening it...")
				openedDoors[o.Name] = o.Position

				err := InteractObject(o, func() bool {
					obj, found := ctx.Data.Objects.FindByID(o.ID)
					return found && !obj.Selectable
				})

				if err != nil {
					ctx.Logger.Debug("Failed to open door", slog.String("error", err.Error()))
				} else {
					utils.Sleep(200)
					return true
				}
			}
		}
	}`

with the the isObjectInPath being:
`func isObjectInPath(dest data.Position, player data.Position, object data.Position, tolerance float64) bool {
	ctx := context.Get()
	minX := math.Min(float64(player.X), float64(dest.X)) - tolerance
	maxX := math.Max(float64(player.X), float64(dest.X)) + tolerance
	minY := math.Min(float64(player.Y), float64(dest.Y)) - tolerance
	maxY := math.Max(float64(player.Y), float64(dest.Y)) + tolerance

	// Check if the object is within the bounds of the line
	if object.X >= int(maxX) || object.X <= int(minX) || object.Y <= int(minY) || object.Y >= int(maxY) {
		return false
	}

	// Check for when player and dest is vertical
	if player.X == dest.X {
		if math.Abs(float64(player.X)-float64(object.X)) >= tolerance {
			ctx.Logger.Debug(fmt.Sprintf("Object is vertical, check failed with value %d ", math.Abs(float64(player.X)-float64(object.X))))
			return false
		}
		return true
	}

	// check for when player and dest is horizontal
	if player.Y == dest.Y {
		if math.Abs(float64(player.Y)-float64(object.Y)) >= tolerance {
			ctx.Logger.Debug(fmt.Sprintf("Object is horizontal, check failed with value %d ", math.Abs(float64(player.Y)-float64(object.Y))))
			return false
		}
		return true
	}

	// check distance of the object from the line between player and dest
	distFromLine := math.Abs(((float64(dest.X)-float64(player.X))*(float64(player.Y)-float64(object.Y)))-((float64(player.X)-float64(object.X))*(float64(dest.Y)-float64(player.Y)))) / math.Sqrt((float64(dest.X)-float64(player.X))*(float64(dest.X)-float64(player.X))+(float64(dest.Y)-float64(player.Y))*(float64(dest.Y)-float64(player.Y)))
	ctx.Logger.Debug(fmt.Sprintf("Object is distance: %d from our path", distFromLine))
	if distFromLine >= tolerance {
		return false
	} else {
		return true
	}
}
`

And from my testing a threshhold of 5 seems to be sufficient.

PR #715, opened 2025-02-28
Arcane Ghost Run

PhoxSillanpaa:
PR creates the option to clear the Arcane Sanctuary, clear it of all ghosts, or just clear the Summoner.


PhoxSillanpaa:
Found a run breaking bug. This run is broken, but I think I know how to fix it.

PhoxSillanpaa:
> Found a run breaking bug. This run is broken, but I think I know how to fix it.

Should be fixed. I ran it for about two hours with no issues.

PR #721, opened 2025-03-03
bugfix: better fix for scheduler validation error?

jasorello:
In https://github.com/hectorgimenez/koolo/pull/706 I offered a fix for the scheduler error that involved setting some defaults in the template.

This is an alternative fix for the problem that doesn't involve adding anything to the template config.yaml.

This is better because some users may not correctly pull in the template when performing a rebuild (or copy their entire config folder into the build dir instead of using `better_build.bat`, like I did at first)
It's also sorta better because there's some nice comments about function in the template config, whereas this scheduler stuff is just kinda noise?

The problem in the code is that the validation code assumes that there are scheduler entries before it starts validating, and previously the list was empty. This PR is another way to ensure that the scheduler config is present before it gets populated and validated. I think this is superior to the prior fix for the reasons I mentioned above.

Alternatively, there's already a fix merged, and maybe this is just unnecessary. :)

### DIFF
--- a/config/template/config.yaml
+++ b/config/template/config.yaml
@@ -13,6 +13,24 @@ closeMiniPanel: false # Set to true to close the mini panel at start of game in 
 hidePortraits: true  # Set to true to hide mercenary and other players portraits (avatar)
 enableCubeRecipes: true # Enable cubing of flawlesses and tokens
 
+scheduler:
+  enabled: false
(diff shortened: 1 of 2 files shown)


artosimonyan:
@jasorello You can leave the defaults in the template, better to have.

Issue #714, opened 2025-02-28
ctx.Data.OpenMenus.LoadingScreen not always accurate when loading between Acts


jasorello:
### Details

I've noticed that in certain run configurations, the bot will not cast buffs as expected at the start of the run.

I added some debug statements and believe there might be additional memory offsets that need to be inspected when determining whether or not a loading screen is currently being viewed.

I can reproduce this 100% on my machine, but I understand other people might have different behavior.

To reproduce:
1. Configure a sorceress to run with a single step, mephisto
2. Put your character in act 4 and then leave the game
3. Start the run, which will move to the WP in pandemonium fortress to act 3 to start the run
4. Notice that the "Post CTA Buffing..." shows in the logs while we're still seeing the "opening door" loading screen
5. When we load into durance of hate, we do not perform any buffs, but we instead immediately start teleporting.

I added a debug statement immediately after the "Post CTA Buffing..." line
debug line: `ctx.Logger.Info(fmt.Sprintf("Context Menus: %v", (ctx.Data.OpenMenus)))`
Log output:
```
time=21:02:57 level=DEBUG msg="Post CTA Buffing..."
time=21:02:57 level=INFO msg="Context Menus: {false false false false false false false false false false false false false false false false false false false false}"
```

It seems that this context struct isn't being populated appropriately for this situation. I'm not familiar enough with go yet to understand how to debug d2go while running the bot, but I assume there needs to be an update over there to handle this? Is there something else going on?

### Version

personal branch built off main

### Logs

```
time=21:02:57 level=DEBUG msg="Post CTA Buffing..."
time=21:02:57 level=INFO msg="Context Menus: {false false false false false false false false false false false false false false false false false false false false}"
```

elobo91:
It should detect it correctly however if loading time are slow it could create issue

Theres a function that is used when creating game. Maybe could solve this issue.
https://github.com/hectorgimenez/koolo/blob/6ab0cd2769d757c2a8475fd1af7c1252a59cd8d8/internal/context/context.go#L166

jasorello:
> It should detect it correctly however if loading time are slow it could create issue
> 
> Theres a function that is used when creating game. Maybe could solve this issue.
> 
> [koolo/internal/context/context.go](https://github.com/hectorgimenez/koolo/blob/6ab0cd2769d757c2a8475fd1af7c1252a59cd8d8/internal/context/context.go#L166)
> 
[...]

PR #722, opened 2025-03-03
buff updates

jasorello:
Originally was just a fix for https://github.com/hectorgimenez/koolo/issues/714

By refreshing the game state before checking for a loading screen, the bot successfully identifies the loading screen and waits for it to complete before running through the Buff routine, avoiding the problem identified in issue 714

Now includes the ability to cast buffs with CTA, support for assassin buffs fade/BoS, and improved identification of CTA and swapping (although it expects the CTA to live on your secondary weapon set)

### DIFF
--- a/internal/action/buff.go
+++ b/internal/action/buff.go
@@ -74,11 +74,21 @@ func Buff() {
 		castSkill(buff)
 	}
 
-	buffCTA([]skill.ID{}) // Future place for buffs to be cast with CTA weapon set
+	hasCTA := ctaFound(*ctx.Data)
+	buffWithCTA := true
+	if hasCTA {
+		if buffWithCTA {
+			buffCTA(ctx.Char.BuffSkills())
+		} else {
+			buffCTA([]skill.ID{})
+		}
+	}
 
 	ctx.Logger.Debug("Post CTA Buffing...")
-	for _, buff := range ctx.Char.BuffSkills() {
-		castSkill(buff)
+	if !hasCTA || !buffWithCTA {
+		for _, buff := range ctx.Char.BuffSkills() {
+			castSkill(buff)
+		}
 	}
 
 	ctx.LastBuffAt = time.Now()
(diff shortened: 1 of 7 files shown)


internal/action/buff.go
```diff
@@ -48,6 +48,7 @@ func Buff() {
 	}
 
 	// Check if we're in loading screen
+	ctx.RefreshGameData()
```
jasorello, line 51: Are there downsides to this besides a few more CPU cycles? Is there a better way?

elobo91: You can selectively update just what you need from GetData()   . 

ctx.GameReader.GetData().OpenMenus.LoadingScreen

Or ctx.GameReader.GetData().OpenMenus
For all OpenMenus

internal/context/context.go
```diff
@@ -164,9 +164,8 @@ func (s *Status) PauseIfNotPriority() {
 	}
 }
 func (ctx *Context) WaitForGameToLoad() {
-	for ctx.Data.OpenMenus.LoadingScreen {
+	for ctx.GameReader.GetData().OpenMenus.LoadingScreen {
 		time.Sleep(100 * time.Millisecond)
-		ctx.RefreshGameData()
```
elobo91, line 169: Hello , can you make sure the removal of the last refreshgamedata after loading screen is completed doesnt create issue with tinymod when creating new game. . Is perfect to only loop the loading screen until loaded but unsure if we dont need one full refresh after to sync. Thanks

jasorello:
I noticed fade wasn't getting cast and poked into why, I think it's because the state for it isn't mapped? I took a stab at a minor refactor here, although I didn't see the state needed for burst of speed.

Presumably this needs to be filled out with all possible buffs and their states

internal/action/buff.go
```diff
@@ -153,7 +177,7 @@ func buffCTA() {
 		ctx.HID.PressKeyBinding(ctx.Data.KeyBindings.MustKBForSkill(skill.BattleCommand))
 		utils.Sleep(180)
 		ctx.HID.Click(game.RightButton, 300, 300)
-		utils.Sleep(100)
```
jasorello, line 156: Is there a way to hook into the player's cast rate here?

internal/action/buff.go
```diff
@@ -103,6 +103,39 @@ func Buff() {
...
+	skill.HolyShield: {state.State(state.Holyshield)},
+	skill.FrozenArmor: {
+		state.State(state.Frozenarmor),
+		state.State(state.Shiverarmor),
+		state.State(state.Chillingarmor),
+	},
+	skill.EnergyShield: {state.State(state.Energyshield)},
+	skill.CycloneArmor: {state.State(state.Cyclonearmor)},
+	skill.Fade:         {state.State(state.Fade)},
+	// skill.BurstOfSpeed: {state.State(state.BurstOfSpeed)}, Not mapped?
```
jasorello, line 117: Where's burst of speed? Did I miss it?

jasorello:
I added some updates based on some discussion with elb, leveraging the ability to check for runewords directly instead of looking for skills on the items, as well as a check for which weapon set we have equipped, instead of again, looking for skills.

jasorello:
I took my own stab at implementing the request implemented in https://github.com/hectorgimenez/koolo/pull/756 and exposed it as a general config item

jasorello:
once https://github.com/hectorgimenez/koolo/pull/761 is merged, this should be updated to leverage the cheaper update calls

PR #723, opened 2025-03-04
Runeword maker

CarlPoppa1:
Added the ability to make runewords whilst leveling.

- It's hardcoded to not make the same runeword in the same tier base. 
- It has a basic base selection routine if there is more than one valid base. You can configure (not in the front-end) whether you want to prioritise damage or defense and whether you want to allow making the runeword in an ethereal base. 
- If you don't want to prioritise damage or defense, the base with the lowest total requirements (required strength + required dex) will be chosen.
- There are currently no equippable checks

This can be the foundation for enabling runeword re-rolling

### DIFF
--- a/internal/action/runeword_recipes.go
+++ b/internal/action/runeword_recipes.go
@@ -1,0 +1,385 @@
+package action
+
+import (
+	"fmt"
+	"slices"
+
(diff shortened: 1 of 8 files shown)


Antizerg:
Cure runeword is very popular for the Act 2 Prayer/Insight combo.

PR #727, opened 2025-03-10
Fix bliz sorc boss killing

artosimonyan:


jasorello:
I had a similar thought and exposed two helper methods to be used across characters: https://github.com/hectorgimenez/koolo/pull/707/files#diff-c95853b0475b178496e3a0a4120e5f77c1e99d2463c010baad27c32ad0e2c15e

PR #731, opened 2025-03-12
Move potions to belt after corpse recovery

Sunblood:
Shift + left clicks on all potions in inventory so they move back into belt slots instead of getting sold.

Requires Shift+click fix from https://github.com/hectorgimenez/koolo/pull/684 to function correctly


artosimonyan:
@Sunblood this should be a part of the belt management script. Also clicking all potions can make unwanted potions go in the belt. 

Check the logic for belt management and implement something similar, only shift+click a pot when needed/has a missing one in the belt

PR #733, opened 2025-03-13
New Gambling (less refresh)

Neo-Silver:
Added 2 new functions to the gambling script to utilize less refresh and safe time.
1) Circlet/Coronet Option
- will basically never refresh and bu coronet turning into circlet and vice versa

2) Less Refresh basic Option
- can handle all sort of user statements to gamble for
- will check each gambler window for ALL stated items
- will just refresh if no stated item is found

To prevent it stuck at for example amulet, that is allways available, it has a map.
Every statement will be counted to 5.
Once it is bought 5 times, it will be excluded until all other items are bought 5 times too.

On sidenote:
- i have both as an option under gambling
- but i think the second function ( aboves 2) ) could be a replacement of the initial gambleitems function

### DIFF
--- a/internal/action/gambling.go
+++ b/internal/action/gambling.go
@@ -48,6 +48,12 @@ func Gamble() error {
 			return errors.New("failed opening gambling window")
 		}
 
+		if ctx.CharacterCfg.Gambling.LessRefresh {
+			return lessRefreshGamble()
+		} else if ctx.CharacterCfg.Gambling.GambleMap {
+			return gambleMap()
+		}
+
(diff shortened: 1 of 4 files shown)


elobo91:
Hi , i think you can remove all force refresh from gambling.go

Inventory is limited to 500ms in d2go anyway so it doesnt help anymore. Only if item is hovered will get updated instant . https://github.com/hectorgimenez/d2go/blob/169a76515285b71e9653678090fb94f7a5743899/pkg/memory/game_reader.go#L60

PR #734, opened 2025-03-14
Fix Den of Evil quest - never starts unless you are in act 1

sorc653:
So if you join the game outside of act 1 you’ll go there before attempting this quest

### DIFF
--- a/internal/run/quests.go
+++ b/internal/run/quests.go
@@ -81,7 +81,12 @@ func (a Quests) Run() error {
 func (a Quests) clearDenQuest() error {
 	a.ctx.Logger.Info("Starting Den of Evil Quest...")
 
-	err := action.MoveToArea(area.BloodMoor)
+	err := action.WayPoint(area.RogueEncampment)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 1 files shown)


Antizerg:
This is awesome.  I found another bug in the Quests.go file if anyone is feeling frisky.

When you run retrieveBookQuest()

Let's assume you don't have any act 3 waypoints, it will get spider forest and great marsh waypoints no problem, but it bugs out after that, I believe due to how Flayer is mapped between Great Marsh and Lower Kurast which bugs out the adjacent area logic (is my theory).

Antizerg:
another update request.  We should add logic so it doesn't run quests that were already completed.  We can steal the logic from the leveling scripts.

I believe this function will open quest log to get the completion:
lvl, _ := s.Data.PlayerUnit.FindStat(stat.Level, 0)

Above example was taken from sorceress_leveling.go.

Then here is an example where we check to see if a quest is completed.

if !a.isCainInTown() && !a.ctx.Data.Quests[quest.Act1TheSearchForCain].Completed() {
		a.deckardCain()
	}

PR #735, opened 2025-03-14
Fix Blizzsorc Pindle & Logging Issue.

xVoidByte:
Fix the GB's of log upon cold immunity of Pindle being met, comment out `KillMonster()` method since it's unused anymore but saved it for future reference if needed.

Blizzsorc will now leave the game if Pindleskin is immune to cold generating a single log, tested on TZ with Nalox with "Cold Immunity Skip" option being on, it ignores the mobs which is the intended behavior.


artosimonyan:
Obsolete with #737

PR #740, opened 2025-03-17
feature: survivability updates for mosaic

jasorello:
While leveling a mosaic build I made some updates, I don't know how much of this would be appealing to the broader public:

Side note: I had trouble getting fade and BO to be cast without updates in https://github.com/hectorgimenez/koolo/pull/722

Updates to mosaic build

1. Don't require all claw skills to be bound; the build will use the attacks it's configured to use as long as there's a keybind for them
2. Switched order of charges to prioritize cobra strike before phoenix. I found this useful when building charge initially so you had a bunch of LL available if you started getting hit more than you expect
3. Related to the above, updated the charge loop to only try to build a specific number of charges at a time
4. Updated to cast cloak of shadows on the target monster as long as that monster doesn't already have cloak of shadows active, or the life tap curse active (in case you have dracul's and don't want to override that curse)
5. Build a single charge of PS for boss killing

TODOs if there's interest:

-  Use of cloak of shadows could be configurable
-  Cloak of shadows should only be attempted to be cast when off cooldown (or maybe only when charge count is low?)
- The number of charge attacks before using a finisher could be configurable

### DIFF
--- a/internal/character/mosaic.go
+++ b/internal/character/mosaic.go
@@ -107,7 +107,7 @@ func (s MosaicSin) buildChargesForSkill(monsterId data.UnitID, skillConfig Charg
 	charges, found := ctx.Data.PlayerUnit.Stats.FindStat(stat.ID(skillConfig.chargeState), 0)
 	attacks := 0
 
-	if !s.MobAlive(monsterId, *s.Data) {
+	if !s.MonsterAliveById(monsterId) {
 		return -1, true
 	}
 
@@ -117,7 +117,7 @@ func (s MosaicSin) buildChargesForSkill(monsterId data.UnitID, skillConfig Charg
 			// Some skills give up to 2 charges per attack
 			plannedAttacks := (neededCharges + skillConfig.chargesPerAttack - 1) / skillConfig.chargesPerAttack
 			attacks += plannedAttacks
-			step.SecondaryAttack(skillConfig.skill, monsterId, plannedAttacks, step.StationaryDistance(1, 4))
+			step.SecondaryAttack(skillConfig.skill, monsterId, plannedAttacks, step.MeleeDistance(2))
 		}
 	}
 
@@ -182,7 +182,7 @@ func (s MosaicSin) AttackLoop(
 			return nil
 		}
 
-		if !s.MobAlive(id, *s.Data) {
+		if !s.MonsterAliveById(id) {
 			return nil
 		}
 
@@ -204,9 +204,8 @@ func (s MosaicSin) AttackLoop(
 			}
 		}
 
-		opts := step.Distance(1, 2)
 		// Finish it off with primary attack
-		step.PrimaryAttack(id, 1, false, opts)
+		step.PrimaryAttack(id, 1, false, step.MeleeDistance(4))
 	}
 }
 
@@ -267,11 +266,6 @@ func (s MosaicSin) KillMonsterSequence(
 		chargeSkillConfig,
 		*ctx,
 	)
-}
-
-func (s MosaicSin) MobAlive(mob data.UnitID, d game.Data) bool {
-	monster, found := s.Data.Monsters.FindByID(mob)
-	return found && monster.Stats[stat.Life] > 0
 }
 
 func (s MosaicSin) BuffSkills() []skill.ID {
--- a/internal/character/mosaic.go
+++ b/internal/character/mosaic.go
@@ -135,7 +135,7 @@ func (s MosaicSin) KillMonsterSequence(
 		// Initial move to monster if we're too far
 		if ctx.PathFinder.DistanceFromMe(monster.Position) > 3 {
 			if s.hasKeyBindingForSkill(skill.DragonFlight) {
-				step.SecondaryAttack(skill.DragonFlight, id, 1)
+				step.SecondaryAttack(skill.DragonFlight, id, 1, step.RangedDistance(0, 18))
 			} else if err := step.MoveTo(monster.Position); err != nil {
 				s.Logger.Debug("Failed to move to monster position", slog.String("error", err.Error()))
 				continue
--- a/internal/character/mosaic.go
+++ b/internal/character/mosaic.go
@@ -28,7 +28,7 @@ type ChargeSkillConfigEntry struct {
 	desiredCharges   int
 }
 
-func (s MosaicSin) getChargeSkillConfig(ctx context.Status) map[skill.ID]ChargeSkillConfigEntry {
+func (s MosaicSin) getBossChargeSkillConfig(ctx context.Status) map[skill.ID]ChargeSkillConfigEntry {
 	return map[skill.ID]ChargeSkillConfigEntry{
 		skill.TigerStrike: {
 			ctx.CharacterCfg.Character.MosaicSin.UseTigerStrike,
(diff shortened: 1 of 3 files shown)


internal/character/mosaic.go
```diff
@@ -87,70 +145,63 @@ func (s MosaicSin) KillMonsterSequence(
...
-				continue
+		// Cobra 3
+		// Phoenix 2
+		// Lightning 3
+		// Tiger 3
+		// Ice 3
+		// Fire 3  // This one breaks life leech, dangerous?
+
+		totalChargeAttacks := 0
+		if ctx.CharacterCfg.Character.MosaicSin.UseCobraStrike && totalChargeAttacks < attacksBeforeKick {
```
jasorello, line 156: What's a more elegant way to write these sections?

internal/character/mosaic.go
```diff
@@ -45,21 +96,26 @@ func (s MosaicSin) KillMonsterSequence(
...
 	lastRefresh := time.Now()
 
+	// TODO: move to config
+	attacksBeforeKick := 4
+	useCloakOfShadows := true
+
+	// Initial cloak of shadows cast for survivability
+	if id, found := monsterSelector(*s.Data); found {
+		if useCloakOfShadows && s.hasKeyBindingForSkill(skill.CloakOfShadows) &&
+			!s.MobHasAnyState(id, []state.State{state.Lifetap, state.CloakOfShadows}) {
```
jasorello, line 106: In the end I was using this with dracul's and I didn't want a cloak cast to overwrite life tap

jasorello:
Mobility has improved a lot with teleport but target acquisition has become way worse. Is it my dragonflight change? Is something else needed?

(I've been testing this branch with the following PRs as well as some of my other dev branches)
```
prs=(
    683 # Fix issue with ping ponging between the two entrances to the same area
    684 # Enable/Repair game.ShiftKey ( shift + left click )
    713 # Walkable logic ( door handling and monsters) 
    716 # Companion Mode + New Flow
)    
    ```

File: internal/character/mosaic.go [hunk 3]
```diff
@@ -45,21 +96,26 @@ func (s MosaicSin) KillMonsterSequence(
...
+			// How do I determine if cloak of shadows is on cooldown?
+			step.SecondaryAttack(skill.CloakOfShadows, id, 1)
+		}
+	}
+
 	for {
 		// Limit refresh rate to 10 times per second to avoid excessive CPU usage
 		if time.Since(lastRefresh) > time.Millisecond*100 {
```
elobo91, line 114: If you are going to rework mosaic i would recommend changing ctx.RefreshGameData() to ctx.GameReader.GetData().Data.PlayerUnit

we only need player states and skills to be precise dont need to update whole data. It will be quicker, cost less cpu and possibly more accurate

jasorello:
I've used with and without enigma. Enigma is great for quicker boss kills but showing up to a boss without charges is painful.  My dream is a hybrid of walking/dragonflight/teleport for murdering everything.

I think that's over my head though. I think the starting point would be to look at how we patrol for monsters and stop telestomping, so that we can acquire an enemy that's slightly outside of melee range and hit it with dragon flight to close the distance. This would help with blizz sorc too. I'd want to make those changes outside of this PR though, maybe pull dragonflight out of it until then as well. (Dragonflight seems to only be used when we're like 4 distance units away which is not what I intended)

internal/character/mosaic.go
jasorello: Looks like we're also using the ctx data to determine monster positioning

```
		// Initial move to monster if we're too far
		if ctx.PathFinder.DistanceFromMe(monster.Position) > 3 {
			if err := step.MoveTo(monster.Position); err != nil {
				s.Logger.Debug("Failed to move to monster position", slog.String("error", err.Error()))
				continue
			}
		}
```
are you suggesting that I update all refs within the killSequence to just update the PlayerUnit and pass that object around instead? will ctx still get updated in the above snippet so that monster pathing can be ok?

elobo91: If you look at bot.go . All data is getting refreshed on a 100ms interval. Only reason why you would force a refresh is if need precise information. You can partially update one little thing while main bot loop keep updating all on 100ms. In this case you need playerunit data only for states/skills ( for charges) . Positioning already handled in move.go and attack go no need to care for this updating

jasorello:
Marked as ready for review, I can add the configurable options in the initial entry if desired, but they seem like generally value adds that don't add time to a run. If you don't want to use em, don't bind em 😎

internal/character/mosaic.go
elobo91: image
so instead of updating everything using this function, you partially update playerunit

jasorello: Ahh, oops. what you offer is much cooler. If all data is refreshed on a 100ms interval, does it add much value to also be refreshing playerUnit on a 100ms interval? Presumably it's here for a reason, but maybe it's not necessary...

elobo91: Everything get updated on a 100ms interval, playerunit included. Only for mosaic, you need to sync otherwize charges might build wrong. Cost way less to only update PlayerUnit on another 100ms loop then having 2 loop on 100ms interval calling fulll refresh data 

This is everything that get updated when calling GetData() ( [refreshgamedata)

](https://github.com/hectorgimenez/d2go/blob/169a76515285b71e9653678090fb94f7a5743899/pkg/memory/game_reader.go#L39)

RobotGaston: thank you for working on improving mosaic combat routine.
I'll give it a spin when I'll have mosaics.

One suggestion if possible.

Mosaic should have special attack sequence on bosses (especially on Diablo and Baal).
Something like cast a static on a boss with a sorc routine but with mosaic assassin spin naturally.

Gathering charges on bosses taking waay to loong and can slow down runs noticeably.
so, different behavior on bosses? - Andariel, Mephisto, Diablo - assassin should not gather all charges, 2 or 3 charges (of any skill - falling meteors for example is enough) are enough before she shall proceed to attack. 

TL;DR

Should ignore long default charge gathering sequence and use a short custom one. But only on bosses.

RobotGaston:
thank you for working on improving mosaic attack routine.

a small suggestion If I may for a consideration if possible : 

there should be a **special** separate combat routine for bosses like Andariel, Diablo and Baal.

Something similar to sorceress _static_ tactic. but with an assassin twist.
Assassin should not gather full stack of charges  when fighting with the boss.
2-3 charges of 1-2 skills is enough to gather before character can stop getting charges and switch to her main attack.
Gather charges to meteors for example, and good to go.
No need to collect them all for a full power.
It takes too long and not needed. Bosses are melting fine under meteors, it's enough for them.

tl dr. 

No need for full charged character on bosses. If we r not getting full charged character - we're saving up time and speeding up our games.

Thank you for improving and changing logic for a class.

jasorello:
There was a comment here (that I can't seem to find) about using less charges/skills for bosses and I agree, that would be nice, I'd just want to figure out an elegant way to get it set up. The challenge is that it reduces survivability, which was the original intent of the PR.

My current assassin has dracul's so the goal on any big target like ubers or bosses is to get meteor and then just slam dragon talon to get life tap cast and then i just left click until they're dead, but without life tap and a lot of life leech that might result in an early death? idk

I'm pondering ways to make it more configurable, maybe a call from killMonsterSequence that gets the order of skills to use and the desired charges to get. This would enable me to say "if we're killing baal, get that meteor charge and go to town"

I also need to update which skills generate 2 charges per attack, like some of the elemental claw skills, because we're wasting some redundant and slow moves building those charges.

LEEJS1:
Why doesn't this use a fade or bos buff?

jasorello:
> Why doesn't this use a fade or bos buff?

As noted above, there's a bug in the buff logic where fade and burst of speed aren't mapped. if you merge in https://github.com/hectorgimenez/koolo/pull/722 the fix for fade is included (i'm not sure what the state for burst of speed is but fade works)

jasorello:
The commit history is a mess on this, let me know if you'd like me to rebase it into a single one, but that would probably wait until https://github.com/hectorgimenez/koolo/pull/761 is merged and can be taken advantage of in here.

PR #716, opened 2025-02-28
Companion Mode + New Flow

artosimonyan:
- Companion mode returns to Koolo
- Completely revamped creating/joining games logic
- Greatly improved online check logic and eliminated a lot of edge cases
- Improved failure to select Character with maximum attempts and a client restart.

PR #748, opened 2025-03-18
Follower Companion

davidfvsilva:
This PR adds a run that makes the character follow the designated leader. The follower will clear around the leader position, open chests, and of course it will follow him into any area. If the follower cannot find the leader, it will go to the town of the act where the leader is, and wait for his TP.

The follower will also do all usual behaviors such as shopping, cubing, stashing, gambling, etc. They also make sure to never run out of keys or scrolls.

I also updated the runs for the leader to send TPs when reaching new areas, in order for followers to follow better (chaos and baal runs are also improved).

The followers can also follow a manually played character.

Enjoy :)

### DIFF
--- a/internal/action/buff.go
+++ b/internal/action/buff.go
@@ -43,10 +43,6 @@ func Buff() {
 	ctx := context.Get()
 	ctx.SetLastAction("Buff")
 
-	if ctx.Data.PlayerUnit.Area.IsTown() || time.Since(ctx.LastBuffAt) < time.Second*30 {
-		return
-	}
-
 	// Check if we're in loading screen
 	if ctx.Data.OpenMenus.LoadingScreen {
 		ctx.Logger.Debug("Loading screen detected. Waiting for game to load before buffing...")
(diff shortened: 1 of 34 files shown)


sogmko:
Follower is a really nice idea, thanks for the works

sogmko:
Can the follower enter the leader's game?

davidfvsilva:
> Can the follower enter the leader's game?

For now you have 2 options: use the companion system @artosimonyan is working on, or manually join a game by pausing the bot. The run can also follow a manually played character.

sogmko:
It works with merge #716 
Thx you

davidfvsilva:
> It works with merge #716 
> Thx you

I've been having tons of fun with this :) enjoy!

sogmko:
It works fine with some little fix, but is there a way that followers target the target of the leader instead of target monters near the leader ?

davidfvsilva:
I wouldn't do that. Many reasons like immunities. They target what they can and clear where they're most effective. If you're doing boss runs where you're not clearing anything else, just make two bots do the same run in the same game.

Antizerg:
In order to follow a manually played character, do you need to use PR #716 with this?

davidfvsilva:
> In order to follow a manually played character, do you need to use PR #716 with this?

I don't think so, you should have a leader name input in the character config: fill it out with your manually played character, pause the bot, manually join the game

PR #749, opened 2025-03-18
Characte Pack - Updated and New characters

davidfvsilva:
Bunch of improved and new characters (and a generic character). Keep in mind that the singer character is best used as a companion using a follower run ;)


Antizerg:
any chance of having hammerdin precasting hammers for throne waves?

davidfvsilva:
That should be part of a baal run rework. Many characters could do it. Probably in the future.

PR #755, opened 2025-03-22
Improve Attack Pattern

Thurys:
_WIP / proposal, input appreciated_

**Current behaviour:**
Right now the koolo gets a list of monsters at the start of each fight and then tries to kill them all. The list is unordered, which leads to the bot teleporting into masses of monsters quite often.

My first commit here shows a quick and dirty band aid fix, as I tried to reduce the number of chickens & deaths while running Diablo with a nova sorc.

**Goal:**
Rewrite the whole attack logic, to include kiting or similar attack strategies for ranged characters, while not worsening the situation for melee.

**Ideas:**
- Have range characters search for safe spots between attacks / when monsters get to close (valid safe spots could be in the range of kill-radius + max-attack-range for example)
- Melee chars could still profit from searching for the closest mob to start attacking, but should not run / teleport during fights or change targets
- Make sure to not leave areas. It's not smart to leave the throne room while kiting.

I'm pretty new to the code base, but I'm sure I can find a way to make the attack strategies better.

Also please note, that the first commit clearly is not what is planned, nor is the ordering of the mobs at the right place, it was just a quick and dirty fix for the time being.

Any help and input is appreciated. Especially in tips on how to best test different characters.

### DIFF
--- a/internal/action/clear_area.go
+++ b/internal/action/clear_area.go
@@ -2,6 +2,7 @@ package action
 
 import (
 	"fmt"
+	"slices"
 
 	"github.com/hectorgimenez/d2go/pkg/data"
 	"github.com/hectorgimenez/koolo/internal/action/step"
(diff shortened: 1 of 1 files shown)


elobo91:
d2go already order units by default

PR #758, opened 2025-03-23
Lightning Sorceress Keybind Update

Neo-Silver:
Updated the Lightning Sorceress Keybind to use the same logic as Blizzard Sorceres for armor buff skills.


SkeletonRocky (commented):
Can confirm it solve the issue!

PR #763, opened 2025-03-25
Automaticly assign best Weapon slot for Pre and Post CTA buffs

Timmynat0r:
Created function to calculate current +allSkills and +classSkills from ID 127 and ID 83
Adjusted buff logic to use this calculation for assigning the "best" weapon slot

Many things still to do, so this is WIP

### DIFF
--- a/internal/action/buff.go
+++ b/internal/action/buff.go
@@ -42,17 +42,17 @@ func BuffIfRequired() {
 func Buff() {
 	ctx := context.Get()
 	ctx.SetLastAction("Buff")
+	ctx.Logger.Debug("WEAPON CACHE CHECKPOINT",
+		slog.Bool("cache_valid", ctx.WeaponBonusCache.IsValid),
+		slog.Any("cache_data", ctx.WeaponBonusCache))
 
 	if ctx.Data.PlayerUnit.Area.IsTown() || time.Since(ctx.LastBuffAt) < time.Second*30 {
 		return
 	}
 
-	// Check if we're in loading screen
 	if ctx.Data.OpenMenus.LoadingScreen {
 		ctx.Logger.Debug("Loading screen detected. Waiting for game to load before buffing...")
 		ctx.WaitForGameToLoad()
-
-		// Give it half a second more
 		utils.Sleep(500)
 	}
 
@@ -68,12 +68,20 @@ func Buff() {
 
 	if len(preKeys) > 0 {
 		ctx.Logger.Debug("PRE CTA Buffing...")
+		if ctx.WeaponBonusCache.IsValid {
+			ctx.Logger.Debug("PRE CTA Buffing with Best Weapon Slot...")
+			bestSlot := getBestWeaponSlot(ctx)
+			if bestSlot != getCurrentWeaponSlot(ctx) {
+				ctx.HID.PressKeyBinding(ctx.Data.KeyBindings.SwapWeapons)
+				utils.Sleep(200)
+			}
+		}
 		for _, kb := range preKeys {
-			utils.Sleep(100)
+			utils.Sleep(200)
 			ctx.HID.PressKeyBinding(kb)
-			utils.Sleep(180)
+			utils.Sleep(280)
 			ctx.HID.Click(game.RightButton, 640, 340)
-			utils.Sleep(100)
+			utils.Sleep(200)
 		}
 	}
 
@@ -92,13 +100,28 @@ func Buff() {
 	if len(postKeys) > 0 {
 		ctx.Logger.Debug("Post CTA Buffing...")
 
-		for _, kb := range postKeys {
-			utils.Sleep(100)
-			ctx.HID.PressKeyBinding(kb)
-			utils.Sleep(180)
-			ctx.HID.Click(game.RightButton, 640, 340)
-			utils.Sleep(100)
+		if ctx.WeaponBonusCache.IsValid {
+			ctx.Logger.Debug("POST CTA Buffing with Best Weapon Slot...")
+			bestSlot := getBestWeaponSlot(ctx)
+			if bestSlot != getCurrentWeaponSlot(ctx) {
+				ctx.HID.PressKeyBinding(ctx.Data.KeyBindings.SwapWeapons)
+				utils.Sleep(200)
+			}
 		}
+
+		for _, kb := range postKeys {
+			utils.Sleep(200)
+			ctx.HID.PressKeyBinding(kb)
+			utils.Sleep(280)
+			ctx.HID.Click(game.RightButton, 640, 340)
+			utils.Sleep(200)
+		}
+		// I have to exclude Berserker from this because "ctx.Data.PlayerUnit.Skills[skill.BattleOrders]" will always be true
+		// and he will not swap weapons to buff righ now
+		if ctx.CharacterCfg.Character.Class != "berserker" {
+			step.SwapToMainWeapon()
+		}
+
 		ctx.LastBuffAt = time.Now()
 	}
 }
@@ -151,11 +174,15 @@ func buffCTA() {
 		}
 
 		ctx.HID.PressKeyBinding(ctx.Data.KeyBindings.MustKBForSkill(skill.BattleCommand))
-		utils.Sleep(180)
+		utils.Sleep(280)
 		ctx.HID.Click(game.RightButton, 300, 300)
-		utils.Sleep(100)
+		utils.Sleep(200)
+		ctx.HID.PressKeyBinding(ctx.Data.KeyBindings.MustKBForSkill(skill.BattleCommand))
+		utils.Sleep(280)
+		ctx.HID.Click(game.RightButton, 300, 300)
+		utils.Sleep(200)
 		ctx.HID.PressKeyBinding(ctx.Data.KeyBindings.MustKBForSkill(skill.BattleOrders))
-		utils.Sleep(180)
+		utils.Sleep(280)
 		ctx.HID.Click(game.RightButton, 300, 300)
 		utils.Sleep(100)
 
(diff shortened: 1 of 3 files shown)


internal/action/buff.go
```diff
@@ -68,12 +68,20 @@ func Buff() {
 
 	if len(preKeys) > 0 {
 		ctx.Logger.Debug("PRE CTA Buffing...")
+		if ctx.WeaponBonusCache.IsValid {
+			ctx.Logger.Debug("PRE CTA Buffing with Best Weapon Slot...")
+			bestSlot := getBestWeaponSlot(ctx)
+			if bestSlot != getCurrentWeaponSlot(ctx) {
+				ctx.HID.PressKeyBinding(ctx.Data.KeyBindings.SwapWeapons)
```
jasorello, line 75: instead of swapping back and forth depending on the skill, you could look at all of the prebuff skills and identify which weapon slot to buff with. Then you'd buff all the skills that benefit from the primary, and then buff all the skills that benefit more from the secondaru

internal/action/buff.go
```diff
@@ -68,12 +68,20 @@ func Buff() {
...
+				utils.Sleep(200)
+			}
+		}
 		for _, kb := range preKeys {
-			utils.Sleep(100)
+			utils.Sleep(200)
```
jasorello, line 80: did you have issues with the lower sleep values?

internal/action/buff.go
```diff
@@ -68,12 +68,20 @@ func Buff() {
 
 	if len(preKeys) > 0 {
 		ctx.Logger.Debug("PRE CTA Buffing...")
+		if ctx.WeaponBonusCache.IsValid {
```
jasorello, line 71: when is this not true?

internal/action/buff.go
```diff
@@ -92,13 +100,28 @@ func Buff() {
...
 			ctx.HID.PressKeyBinding(kb)
-			utils.Sleep(180)
+			utils.Sleep(280)
 			ctx.HID.Click(game.RightButton, 640, 340)
-			utils.Sleep(100)
+			utils.Sleep(200)
 		}
+		// I have to exclude Berserker from this because "ctx.Data.PlayerUnit.Skills[skill.BattleOrders]" will always be true
+		// and he will not swap weapons to buff righ now
+		if ctx.CharacterCfg.Character.Class != "berserker" {
```
jasorello, line 121: I think you know this, but switching to the logic that just looks at weaponslot 0 or 1 would allow you to avoid this issue

internal/action/buff.go
```diff
@@ -176,3 +203,104 @@ func ctaFound(d game.Data) bool {
...
+	)
+
+	if !hasCTA {
+		ctx.Logger.Debug("CTA not found - cache marked invalid")
+	}
+}
+
+// Here we calculate all the bonuses from the player's gear and skills
+// and return the total bonus for the All +Skills (ID 127) and +AddClassSkills (ID 83) stats.
+// Right now we ignore +SpecificSkills from Stat ID 97, until we build a UseCase for it.
```
jasorello, line 250: doing specific skills will be such a nightmare, lmao

internal/action/buff.go
```diff
@@ -176,3 +203,104 @@ func ctaFound(d game.Data) bool {
...
+	for layer := 0; layer < 10; layer++ {
+		if s, found := ctx.Data.PlayerUnit.Stats.FindStat(stat.AllSkills, layer); found {
+			allSkills += s.Value
+		}
+	}
+	total += allSkills
+
+	classSkills := 0
+	for layer := 0; layer < 10; layer++ {
+		if s, found := ctx.Data.PlayerUnit.Stats.FindStat(stat.AddClassSkills, layer); found {
```
jasorello, line 264: is this one smartly looking for the correct AddClassSkills stat based on our current class? Or will a sorc wearing treachery have a +2 assassin skills value show up?

internal/action/buff.go
jasorello: nevermind, i assumed you were looking at specific +skills as well, this is moot.

internal/action/buff.go
```diff
@@ -176,3 +203,104 @@ func ctaFound(d game.Data) bool {
...
+	} else {
+		ctx.Logger.Debug("Selected slot 2 as best weapon slot",
+			slog.Int("slot1_bonus", ctx.WeaponBonusCache.Slot1AllClassBonus),
+			slog.Int("slot2_bonus", ctx.WeaponBonusCache.Slot2AllClassBonus),
+		)
+		return 2
+	}
+}
+
+func getCurrentWeaponSlot(ctx *context.Status) int {
```
jasorello, line 301: as noted above, you can avoid this by just explicitly using a weapon slot from `ctx.Data.ActiveWeaponSlot`

internal/action/buff.go
Timmynat0r: will implement this next

internal/action/buff.go
Timmynat0r: as stated above, will go down that route in next iteration

internal/action/buff.go
Timmynat0r: should be resolved

internal/action/buff.go
Timmynat0r: implemented, so now works for all classes

internal/action/buff.go
Timmynat0r: kept ctx.WeaponBonusCache.IsValid in case of error while calculating GearCache, so we can always fall back on original buff logic

internal/action/buff.go
Timmynat0r: implemented, so now works for all classes

internal/action/buff.go
Timmynat0r: V3 should be able to do this right now

Timmynat0r:
ctx.Data.PlayerUnit.Stats.FindStat(stat.AddSkillTab, tabskills)

This already gives me a predetermined list:

	188: {
		0:  "+# to Bow and Crossbow Skills (Amazon only)",     // bowandcrossbowskilltab
		1:  "+# to Passive and Magic Skills (Amazon only)",    // passiveandmagicskilltab
		2:  "+# to Javelin and Spears Skills (Amazon only)",   // javelinandspearskilltab
		8:  "+# to Fire Skills (Sorceress only)",              // fireskilltab
		9:  "+# to Lightning Skills (Sorceress only)",         // lightningskilltab
		10: "+# to Cold Skills (Sorceress only)",              // coldskilltab
		16: "+# to Curses Skills (Necromancer only)",          // cursesskilltab
		17: "+# to Poison and Bone Skills (Necromancer only)", // poisonandboneskilltab
		18: "+# to Summoning Skills (Necromancer only)",       // necromancersummoningskilltab
		24: "+# to Paladin Combat Skills (Paladin only)",      // palicombatskilltab
		25: "+# to Offensive Aura Skills (Paladin only)",      // offensiveaurasskilltab
		26: "+# to Defensive Aura Skills (Paladin only)",      // defensiveaurasskilltab
		32: "+# to Barbarian Combat Skills (Barbarian only)",  // barbcombatskilltab
		33: "+# to Mastery Skills (Barbarian only)",           // masteryesskilltab
		34: "+# to War Cry Skills (Barbarian only)",           // warcriesskilltab
		40: "+# to Druid Summoning Skills (Druid only)",       // druidsummoningskilltab
		41: "+# to Shapeshifting Skills (Druid only)",         // shapeshiftingskilltab
		42: "+# to Elemental Skills (Druid only)",             // elementalskilltab
		48: "+# to Traps Skills (Assassin only)",              // trapsskilltab
		49: "+# to Shadow Discipline Skills (Assassin only)",  // shadowdisciplinesskilltab
		50: "+# to Martial Arts Skills (Assassin only)",       // martialartsskilltab
		
My problem right now is that those "layers" are not connected to skills.go (atleast not that i know of)
So at this stage, yes i can read those +skills, their value, and what they should influence.

So for example:
I have a weapon in slot1, that gives +2 to Cold Skills
ctx.Data.PlayerUnit.Stats.FindStat(stat.AddSkillTab, tabskills)
would give me ID 188 val=2 tabskills=10

and my BuffSkills() has FrozenArmor [skill.id] 40 and no other informations

Now i need a "connection" from [skill.id] 40 to tabskills=10

Whats the easiest way to do that ?
The only thing i have in my head right now is to generate a static map for this.
Or am i missing something ?

CarlPoppa1:
Why not do something like this?
```
	buff := skill.FrozenArmor
	tabskillLayer := (int((buff)-6)/30)*8 + (buff.Desc().Page - 1)
	ctx.Logger.Debug(fmt.Sprintf("tabskillLayer: %d", tabskillLayer))
```

CarlPoppa1:
Sorry I don't understand what you mean. My example takes 40 and returns 10, isn't that what is needed?

Timmynat0r:
you have 5 different kinds of "stats" as +skills that you can have

+skills to All Skills (example: ansihinilus +1allskills)
+skills to Class Skills (example: tal amu +2sorc skills)
+skills to singleskills (example: snowclash +2 To Chilling Armor etc.)
+skills to nonClassSkills (example: CTA +6 to BattleOrders)
+skills to Skill Tree Tabs (example: Bo Sticks +3 warcries)

and the last example is not mapped in skills.go in d2go right now
this would be 
ctx.Data.PlayerUnit.Stats.FindStat(stat.AddSkillTab, tab) ID188 value:3 tab: 34

maybe i just overcomplicate stuff, but if i want to really calculate the best slot for every possibility, i would need this

If i had this, a BO Barb script, would automaticly use +3warcries sticks in the second slot, and cast his bo/bc

CarlPoppa1:
Nah it sounds good, but I don't see why my suggestion wouldn't work, this is untested but it seems exactly what's needed, sorry if I'm being stupid:

```
func getSpecificSkillBonuses(ctx *context.Status, skillIDs []skill.ID) map[skill.ID]int {
	bonuses := make(map[skill.ID]int)

	for _, skillID := range skillIDs {
		**tabskillLayer := (int((skillID)-6)/30)*8 + (skillID.Desc().Page - 1)**
		if s, found := ctx.Data.PlayerUnit.Stats.FindStat(stat.SingleSkill, int(skillID)); found {
			bonuses[skillID] = s.Value
		} else if s, found := ctx.Data.PlayerUnit.Stats.FindStat(stat.NonClassSkill, int(skillID)); found {
			bonuses[skillID] = s.Value
		} **else if s, found := ctx.Data.PlayerUnit.Stats.FindStat(stat.AddSkillTab, tabskillLayer); found {
			bonuses[skillID] = s.Value**

		}

	}
	return bonuses
}
```

Timmynat0r:
> Why not do something like this?
> 
> ```
> 	buff := skill.FrozenArmor
> 	tabskillLayer := (int((buff)-6)/30)*8 + (buff.Desc().Page - 1)
> 	ctx.Logger.Debug(fmt.Sprintf("tabskillLayer: %d", tabskillLayer))
> ```
> 
> 6 is the number of non-class skills at the start of skills.go. 30 is the number of skills for each class

It took some time and another pair of eyes to get what you ment here.
I even almost succesfully used this here, BUT for this to work, right now, we need the list i now added at the end, because Druid Skills and Assassin Skills are missing in "github.com/hectorgimenez/d2go/pkg/data/skill/skilldesc.go"

And since the skillIDs for Assasin and Druid skills, are much higher, even with the Information for the Page, your calculation propably wouldnt have worked.

buff := skill.Tornado [245]
tabskillLayer := (int((buff)-6)/30)*8 + (buff.Desc().Page =-1) -->page propably 3, but not defined in skilldesc.go
tabskillLayer =58

But the tabskill Layer is 42 for elemental (druid)

PR #771, opened 2025-04-02
Avoid (ban) signature detection by Blizzard

xVoidByte:
The PR contains some security changes:
- Koolo binaries will be now obfuscated upon build.
- Unique hash per binary.
- Randomized name to prevent signature/process detection.
- Updated `better_build.exe` along with the documentation and other files.

Users will be required to do small modifications prior to current build as written in the documentation.

Two testers have been running for a few hours now without any issues (awaiting third person to give results tomorrow).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,18 +11,35 @@ jobs:
     steps:
       - name: "Checkout"
         uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


filipesarturi:
Nice job! Are you still using it without any problems?

